PR #9, opened 2014-12-18
XD-2357-2391 Initial implementation of Kafka ChannelAdapter using the SimpleConsumer API

mbogoevici:
XD-2357-2391 Initial implementation of Kafka ChannelAdapter using the SimpleConsumer API

The implementation has the following components:

- KafkaConnectionFactory - manages and caches connections to a set of brokers
- KafkaBrokerConnection - handles low-level SimpleConsumer API calls and converts them to internal objects such as KafkaMessage and KafkaMessageBus
- KafkaTemplate - implements higher-level read operations on a Kafka broker
- OffsetManager - stores (potentially in a persistent fashion) offsets for a group of consumers, configures/resets initial offsets as necessary - can be configured to start at an arbitrary offset in a partition (e.g. replay since offset 100), or relative to a given timestamp (e.g. replay since Monday)
- KafkaMessageListenerContainer - retrieves messages from a given broker for an arbitrary set of partitions or topics (i.e. all partitions in the topics), invoking a MessageListener. Concurrency is adjustable, and allows processing multiple sets of partitions in parallel (while preserving ordering within a partition). Can poll multiple brokers (each on a parallel thread)
- AbstractDecodingMessageListener - utility base class for a MessageListener implementation that decodes the payload and key
- KafkaInboundChannelAdapter delegating to the KafkaMessageListenerContainer

Tests:
- Single and multi-broker configurations
- Partition subset retrieval
- Replicated sets
- Compression

TO DO:
- add tests for arbitrary start offsets
- add support for leader change failover
- add support for (optionally) retrieving configuration/leader information from Zookeeper
- namespace support

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,6 +18,7 @@ sourceCompatibility = targetCompatibility = 1.6
 
 ext {
 	avroVersion = '1.7.6'
+	gsCollectionsVersion = '5.0.0'
 	jacocoVersion = '0.7.0.201403182114'
 	kafkaVersion = '0.8.1.1'
 	metricsVersion = '2.2.0'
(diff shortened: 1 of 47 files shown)


mbogoevici:
Comments addressed. Withdrawn in favour of https://github.com/spring-projects/spring-integration-kafka/pull/12

PR #14, opened 2015-01-23
INTEXT-129: Namespace support for `KafkaMDCA`

artembilan:
JIRA: https://jira.spring.io/browse/INTEXT-129

### DIFF
--- a/src/main/resources/org/springframework/integration/config/xml/spring-integration-kafka-1.0.xsd
+++ b/src/main/resources/org/springframework/integration/config/xml/spring-integration-kafka-1.0.xsd
@@ -492,4 +492,182 @@
 		</xsd:complexType>
 	</xsd:element>
 
+	<xsd:element name="message-driven-channel-adapter">
+		<xsd:annotation>
+			<xsd:documentation>
+				The definition for the Spring Integration Kafka
+				KafkaMessageDrivenChannelAdapter.
+			</xsd:documentation>
+		</xsd:annotation>
+		<xsd:complexType>
+			<xsd:attributeGroup ref="integration:channelAdapterAttributes"/>
+			<xsd:attribute name="send-timeout" type="xsd:string">
+				<xsd:annotation>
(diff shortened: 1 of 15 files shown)


artembilan:
@mbogoevici , @garyrussell , let me know if there is something else to cover with Namespace support.

With this Exstension I see further tasks:
1. Docs for new feature
2. Spring Boot Autoconfiguration 
3. Java DSL Namespace Factory

WDYT?

garyrussell:
Sounds good to me; although maybe only 1. is __required__ for 1.0.0.GA ?

I will review this on the 'plane.

mbogoevici:
Sounds great to me, as well.

src/main/resources/org/springframework/integration/config/xml/spring-integration-kafka-1.0.xsd
```diff
@@ -492,4 +492,182 @@
...
+					<xsd:documentation><![CDATA[
+						Allows you to specify how long this inbound-channel-adapter
+						will wait for the message (containing the retrieved entities)
+						to be sent successfully to the message channel, before throwing
+						an exception.
+
+						Keep in mind that when sending to a DirectChannel, the
+						invocation will occur in the sender's thread so the failing
+						of the send operation may be caused by other components
+						further downstream. By default the Inbound Channel Adapter
```
garyrussell: ? The usual verbiage here is something like "Only applies if the MessageChannel can block - such as a bounded QueueChannel that is full".

artembilan:
Thanks, @garyrussell  - fixed. It was really a stupid copy/paste from the code above in this XSD :smile: .
Also added docs on the matter.

garyrussell:
LGTM; minor doc polish here: https://github.com/garyrussell/spring-integration-kafka/commit/86d874747c940e27000bf2c4c1030620c11c09be

garyrussell:
Merged as 37297389d504a031e2adbf5ab9dae4f7b430e766

PR #15, opened 2015-01-30
INTEXT-131-134: Add deletion support to OffsetManagers and add a Kafka Topic based OffsetManager

mbogoevici:
Add deletion support to OffsetManagers and add a Kafka Topic based OffsetManager

- Separate MetadataStore-related functionality and extract an AbstractOffsetManager to handle resetting/retrieving default offsets from Kafka
- Move connection parameters to the Configuration class
- Add deletion support to OffsetManagers
- Introduce KafkaOffsetManager


src/main/java/org/springframework/integration/kafka/core/DefaultConnectionFactory.java
```diff
@@ -227,14 +162,48 @@ public void refreshLeaders(Collection<String> topics) {
...
-		return partitionBrokerMapReference.get();
+	private MetadataCache getMetadataCache() {
+		return metadataCacheHolder.get();
+	}
+
+	@SuppressWarnings("serial")
+	private static class ErrorlessTopicMetadataPredicate implements Predicate<TopicMetadata> {
+		@Override
+		public boolean accept(TopicMetadata topicMetadata) {
+			return topicMetadata.errorCode() == ErrorMapping.NoError();
```
artembilan: Minor. Flip the condition in case of `topicMetadata` is `null`

src/main/java/org/springframework/integration/kafka/core/DefaultConnectionFactory.java
```diff
@@ -50,28 +51,21 @@
...
+	public static final Predicate<TopicMetadata> errorlessTopicMetadataPredicate = new ErrorlessTopicMetadataPredicate();
+
 	private final GetBrokersByPartitionFunction getBrokersByPartitionFunction = new GetBrokersByPartitionFunction();
 
 	private final ConnectionInstantiationFunction connectionInstantiationFunction = new ConnectionInstantiationFunction();
 
 	private final Configuration configuration;
 
-	private final AtomicReference<PartitionBrokerMap> partitionBrokerMapReference = new AtomicReference<PartitionBrokerMap>();
+	private final AtomicReference<MetadataCache> metadataCacheHolder =
```
artembilan: Why can't avoid `AtomicReference` here?
However I guess: to avoid `volatile` for this class, becuase `MetadataCache` is immutable.
OK. let it be.

src/main/java/org/springframework/integration/kafka/core/DefaultConnectionFactory.java
```diff
@@ -172,12 +103,17 @@ public void setFetchMetadataTimeout(int fetchMetadataTimeout) {
 	 */
 	@Override
 	public BrokerAddress getLeader(Partition partition) {
-		try {
-			lock.readLock().lock();
```
artembilan: Why `readLock` isn't needed anymore? Becuase need to upgrade to `writeLock` in `refreshLeaders` ? :smile: 
Anyway `getMetadataCache()` would be better for readability

src/main/java/org/springframework/integration/kafka/listener/AbstractOffsetManager.java
```diff
@@ -0,0 +1,136 @@
...
+
+	protected long referenceTimestamp = KafkaConsumerDefaults.DEFAULT_OFFSET_RESET;
+
+	protected ConnectionFactory connectionFactory;
+
+	protected Map<Partition, Long> initialOffsets;
+
+	public AbstractOffsetManager(ConnectionFactory connectionFactory, Map<Partition, Long> initialOffsets) {
+		this.connectionFactory = connectionFactory;
+		this.initialOffsets = initialOffsets == null ? new HashMap<Partition, Long>() : initialOffsets;
```
artembilan: Can't we extract one more ctor without `initialOffsets` ?

src/main/java/org/springframework/integration/kafka/listener/AbstractOffsetManager.java
```diff
@@ -0,0 +1,136 @@
...
+	 * {@link MetadataStore}.
+	 *
+	 * @param consumerId the consumer ID
+	 */
+	public void setConsumerId(String consumerId) {
+		this.consumerId = consumerId;
+	}
+
+	/**
+	 * A timestamp to be used for resetting initial offsets, if they are not available in the {@link MetadataStore}
```
artembilan: Mentioning of `MetadataStore` is wrong here.

src/main/java/org/springframework/integration/kafka/listener/AbstractOffsetManager.java
```diff
@@ -0,0 +1,136 @@
...
+		this.initialOffsets = initialOffsets == null ? new HashMap<Partition, Long>() : initialOffsets;
+	}
+
+	public String getConsumerId() {
+		return consumerId;
+	}
+
+	/**
+	 * The identifier of a consumer of Kafka messages. Allows to store separate sets of offsets in the
+	 * {@link MetadataStore}.
```
artembilan: DITO

src/main/java/org/springframework/integration/kafka/listener/AbstractOffsetManager.java
```diff
@@ -0,0 +1,136 @@
...
+	protected String consumerId = KafkaConsumerDefaults.GROUP_ID;
+
+	protected long referenceTimestamp = KafkaConsumerDefaults.DEFAULT_OFFSET_RESET;
+
+	protected ConnectionFactory connectionFactory;
+
+	protected Map<Partition, Long> initialOffsets;
+
+	public AbstractOffsetManager(ConnectionFactory connectionFactory, Map<Partition, Long> initialOffsets) {
+		this.connectionFactory = connectionFactory;
```
artembilan: `Assert.notNull(connectionFactory)`

src/main/java/org/springframework/integration/kafka/listener/KafkaTopicOffsetManager.java
```diff
@@ -0,0 +1,505 @@
...
+	}
+
+	@Override
+	public void flush() throws IOException {
+		// not supported
+	}
+
+	@Override
+	public void close() throws IOException {
+		producer.close();
```
artembilan: Shouldn't this class be `DisposableBean` or Spring takes care about `Closeable` well ?

src/main/java/org/springframework/integration/kafka/util/TopicUtils.java
```diff
@@ -0,0 +1,158 @@
...
+					}
+					for (PartitionMetadata partitionMetadata : partitionMetadatas) {
+						if (partitionMetadata.errorCode() != ErrorMapping.NoError()) {
+							throw (Exception) ErrorMapping.exceptionFor(partitionMetadata.errorCode());
+						}
+					}
+					return topicMetadata;
+				}
+			});
+			// work around an issue in Spring
```
artembilan: Which issue ?

src/main/java/org/springframework/integration/kafka/util/TopicUtils.java
```diff
@@ -0,0 +1,158 @@
...
+
+			return topicMetadata;
+		}
+		catch (Exception e) {
+			log.error("Cannot initialize MessageBus", e);
+			throw new RuntimeException("Cannot initialize message bus:", e);
```
artembilan: Something wrong is here. I mean `message bus` words

artembilan:
@mbogoevici , that's all. I see some few redundant `@SupressWarning`, but it is minor and we can live that. As well as all those my comments can be fixed on merge.

Just give some answers and let's see feedback from @garyrussell if he has a wish to review too!

src/main/java/org/springframework/integration/kafka/core/DefaultConnectionFactory.java
mbogoevici: I don't think that flipping would help, unless we box and use `equals`. But seems very defensive, seeing as `topicMetadata` can't really be `null`, based on its use.

src/main/java/org/springframework/integration/kafka/core/DefaultConnectionFactory.java
mbogoevici: `volatile` could work as well, eventually. I think it communicates better the intent of being a data holder, though.

src/main/java/org/springframework/integration/kafka/core/DefaultConnectionFactory.java
mbogoevici: Yes, but `ReentrantReadWriteLock` does not support write upgrades. 

What do you mean with the comment about `getMetadataCache()` ?

src/main/java/org/springframework/integration/kafka/listener/AbstractOffsetManager.java
mbogoevici: I suppose it could. I never had a use for it, quite frankly, though.

src/main/java/org/springframework/integration/kafka/listener/KafkaTopicOffsetManager.java
mbogoevici: No, Spring does not, but `KafkaMessageListenerContainer` does on is own `stop`. I see how `DisposableBean` would be useful, though.

src/main/java/org/springframework/integration/kafka/util/TopicUtils.java
mbogoevici: Missed that comment.

src/main/java/org/springframework/integration/kafka/core/DefaultConnectionFactory.java
artembilan: ???
````java
 *       // Must release read lock before acquiring write lock
 *       rwl.readLock().unlock();
 *       rwl.writeLock().lock();
 *       try {
 *         // Recheck state because another thread might have
 *         // acquired write lock and changed state before we did.
 *         if (!cacheValid) {
 *           data = ...
 *           cacheValid = true;
 *         }
 *         // Downgrade by acquiring read lock before releasing write lock
 *         rwl.readLock().lock();
 *       } finally {
 *         rwl.writeLock().unlock(); // Unlock write, still hold read
 *       }
````

I just mean that you have already `getMetadataCache()` instead of that `metadataCacheHolder.get()`. OK, nevermind this case. 

But `readLock` concern is still here...

File: src/main/java/org/springframework/integration/kafka/util/TopicUtils.java (hunk 10 above)
Comment (mbogoevici) [inline src/main/java/org/springframework/integration/kafka/util/TopicUtils.java]: Yes, that's slipped when I've ported it back from the `KafkaMessageBus` in XD.

File: src/main/java/org/springframework/integration/kafka/core/DefaultConnectionFactory.java (hunk 3 above)
Comment (mbogoevici) [inline src/main/java/org/springframework/integration/kafka/core/DefaultConnectionFactory.java]: I see your point now, @artembilan - I misread your comment (for whatever reason I thought it'd referred to `getPartitions`, that uses a read lock - no reason why same pattern shouldn't be applied to `getLeaders` too.

**Comment by mbogoevici** [comment]
@artembilan I've added my fixes in - thanks for the review - let me know if there's anything left.

File: src/main/java/org/springframework/integration/kafka/core/DefaultConnectionFactory.java [hunk 11]
```diff
@@ -172,13 +103,32 @@ public void setFetchMetadataTimeout(int fetchMetadataTimeout) {
...
+				leader = getMetadataCache().getLeader(partition);
+			}
 		}
 		finally {
 			lock.readLock().unlock();
 		}
+		if (leader == null) {
+			try {
+				lock.writeLock().lock();
+				this.refreshMetadata(Collections.singleton(partition.getTopic()));
```
artembilan: I think we should have `synchronized double check` here to avoid one more hard extra refresh.
If we wait here for the lock, another `writer` could refresh already, so our `leader` could appear here and there won't be reason to `refresh` on more time.

src/main/java/org/springframework/integration/kafka/core/DefaultConnectionFactory.java
```diff
@@ -227,14 +176,48 @@ public void refreshLeaders(Collection<String> topics) {
...
+		}
+		finally {
+			lock.readLock().unlock();
+		}
+		// if we got here, it means that the data was not available, we should try a refresh. The lock is reentrant
+		// so we will not block ourselves
+		if (returnedPartitions == null) {
+			try {
+				lock.writeLock().lock();
+				this.refreshMetadata(Collections.singleton(topic));
```
artembilan: DITO

src/main/java/org/springframework/integration/kafka/listener/KafkaTopicOffsetManager.java
```diff
@@ -0,0 +1,505 @@
...
+
+/**
+ * Implementation of an {@link OffsetManager} that uses a Kafka topic as the underlying support.
+ * For its proper functioning, the Kafka server(s) must set {@code log.cleaner.enable=true}. It relies on the property
+ * {@code cleanup.policy=compact} to be set on the target topic, and if the topic is not found,
+ * it will create a topic with the appropriate settings.
+ *
+ * @author Marius Bogoevici
+ */
+@SuppressWarnings({"unchecked", "rawtypes"})
```
artembilan: As I said before: this `@SuppressWarnings` is redundant. My Gradle doesn't complain, when I comment it

src/main/java/org/springframework/integration/kafka/listener/KafkaTopicOffsetManager.java
```diff
@@ -0,0 +1,505 @@
...
+public class KafkaTopicOffsetManager extends AbstractOffsetManager implements InitializingBean {
```
artembilan: We talked with you about `DisposableBean`, but you haven't addressed it.
Have I missed anything ?
I see that `KafkaMessageListenerContainer` closes `offsetManager` on its `stop()`. But for me it looks not kosher, because we can share the same `offsetManager` between different containers/adapters. So, the `stop` of this one may cause an issue - `producer.close()`.
From other side we can `stop/start` container/adapter any time. And having this `producer.close()` we become into inconsistent state of our system.

So, it's really a bug IMO

src/main/java/org/springframework/integration/kafka/util/TopicUtils.java
```diff
@@ -0,0 +1,157 @@
...
+					}
+					return topicMetadata;
+				}
+			});
+			return topicMetadata;
+		}
+		catch (Exception e) {
+			String message = String.format("Cannot retrieve metadata for topic '%s'", topicName);
+			log.error(message, e);
+			throw new RuntimeException(message,e);
```
artembilan: Isn't that `log.error` redundant in case of `throw`?
From other side can't we throw more specific exception here?

src/main/java/org/springframework/integration/kafka/listener/KafkaTopicOffsetManager.java
mbogoevici: Okay. I see your point. I've never actually intended to have OffsetManager shared between multiple instances KafkaMessageListenerContainer scenarios (I can't imagine a practical scenario where we would actually want that), but, in theory, you're right. I see no harm in eventually going your route.

artembilan: One more time: even if we don't share `OffsetManager` we can start/stop container many times. But with this `producer.close()` we don't allow to restart container. True ?

mbogoevici: Again, I don't really see it happening in practice. But you're right that we don't need to have that coupling.

src/main/java/org/springframework/integration/kafka/core/DefaultConnectionFactory.java
mbogoevici: Yeah, we can do that. Again, based on use it's a very unlikely situation - but we can skip that refresh here.

src/main/java/org/springframework/integration/kafka/listener/KafkaTopicOffsetManager.java
artembilan: In practice? it is typical SI application behaviour, when we manage our `Lifecycles`. ControlBus is standard component to do that:

````xml
<control-bus channel="controlBusChannel"/>
````
````java
this.controlBusChannel.send(new GenericMessage<String>("@myKafkaAdapter.stop()"));
````

Comment (mbogoevici) [inline src/main/java/org/springframework/integration/kafka/listener/KafkaTopicOffsetManager.java]: Well, I mentioned I'll change that already.

File: src/main/java/org/springframework/integration/kafka/util/TopicUtils.java (hunk 15 above)
Comment (mbogoevici) [inline src/main/java/org/springframework/integration/kafka/util/TopicUtils.java]: Well, it may be redundant, but I'd rather have this logged here, in case the exception is swallowed. This is a serious error condition.

**Comment by mbogoevici** [comment]
Updated.

**Comment by mbogoevici** [comment]
Updated.

File: src/main/java/org/springframework/integration/kafka/listener/KafkaTopicOffsetManager.java [hunk 16]
```diff
@@ -0,0 +1,505 @@
...
+	private int replicationFactor;
+
+	private int maxBatchSize = 200;
+
+	private boolean batchWrites = true;
+
+	private int requiredAcks = 1;
+
+	public KafkaTopicOffsetManager(ZookeeperConnect zookeeperConnect,
+			ConnectionFactory connectionFactory, String topic, Map<Partition, Long> initialOffsets) {
```
mbogoevici: I've been struggling with that problem for a while. On one hand, there's the inconsistency. On the other hand, I'd prefer to use the ConnectionFactory and reuse the cached Connection. However the point is somehow moot, as producers (who do the more work) have their own separate connections anyway.

mbogoevici: So we could perhaps simplify the constructor.

mbogoevici: Also we can close connectionFactory after reading offset data, and save a socket for future ;).

artembilan:
@mbogoevici , here is a polishing https://github.com/artembilan/spring-integration-kafka/commit/c90cf86c5765da7d94efe78b394a9eee335e405c
See the commit comments

mbogoevici:
@artembilan, I've added a comment on the `KafkaTopicOffsetManager` changes

artembilan:
Merged as https://github.com/spring-projects/spring-integration-kafka/commit/a38cbca7c9a47af6ba515c4e7e15b5c2751e4a0d

PR #16, opened 2015-02-05
Log metadata fetch failures at INFO level. We do not treat them as fatal.

mbogoevici:
....


garyrussell:
Merged as ab6f0ee6a962141f7056b5c9a91c1aca6731a3d3

Reduced commit comment headline to <= 50.

Fixed newline in gradle.properties.

PR #20, opened 2015-03-04
INTEXT-138,143 Add support for manual acknowledgment of messages

mbogoevici:
- Introduce an Acknowledgment object that message processors can invoke;
- Introduce an AcknowedgingMessageListener variation of the MessageListener that receives an Acknowledgment reference for the processed message;
- Add 'autoCommitOffset' settings to the KafkaMessageListenerContainer and KafkaMessageDrivenChannelAdapter, and the ability to inject a MessageListener or an AcknowledgingMessageListener in either (the allowed type depending on the offset management strategy)
- Prepopulate a message header for SI messages created by the KafkaMessageDrivenChannelAdapter if autoCommit is enabled or disabled;
- remove extra `_` from KafkaHeaders
- OffsetManager only sets values that are higher than the ones already set in a session (barring reset) - this is to prevent asynchrous acks to mistakenly revert checkpoints

### DIFF
--- a/src/main/java/org/springframework/integration/kafka/listener/DefaultAcknowdledgment.java
+++ b/src/main/java/org/springframework/integration/kafka/listener/DefaultAcknowdledgment.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.kafka.listener;
+
+import org.springframework.integration.kafka.core.KafkaMessage;
+import org.springframework.integration.kafka.core.Partition;
+
+/**
+ * Default implementation for an {@link Acknowledgment} that defers to an underlying
+ * {@link OffsetManager}.
+ *
+ * @author Marius Bogoevici
+ */
+public class DefaultAcknowdledgment implements Acknowledgment {
+
+	private final OffsetManager offsetManager;
+
+	private final Partition partition;
+
+	private final Long offset;
+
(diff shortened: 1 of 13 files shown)


src/main/java/org/springframework/integration/kafka/listener/DefaultAcknowdledgment.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.integration.kafka.core.KafkaMessage;
+import org.springframework.integration.kafka.core.Partition;
+
+/**
+ * Default implementation for an {@link Acknowledgment} that defers to an underlying
+ * {@link OffsetManager}.
+ *
+ * @author Marius Bogoevici
+ */
+public class DefaultAcknowdledgment implements Acknowledgment {
```
garyrussell: typo

src/main/java/org/springframework/integration/kafka/support/KafkaHeaders.java
```diff
@@ -28,8 +29,12 @@
 
 	public static final String MESSAGE_KEY = PREFIX + "messageKey";
 
-	public static final String PARTITION_ID = PREFIX + "_partitionId";
+	public static final String PARTITION_ID = PREFIX + "partitionId";
 
-	public static final String OFFSET = PREFIX + "_offset";
+	public static final String OFFSET = PREFIX + "offset";
```
garyrussell: Are we going back to plan A (add this to 1.1.x only).

If not, unfortunately, we can't change these in 1.0.x (breaking change).

src/main/java/org/springframework/integration/kafka/listener/DefaultAcknowdledgment.java
mbogoevici: D'oh.

src/main/java/org/springframework/integration/kafka/support/KafkaHeaders.java
mbogoevici: To be honest, I'd move to 1.1 and get done with it at all. This and retriability are big enough features to warrant a 1.1 I believe.

mbogoevici: Probably the best idea is to segregate that fix and have manual ack supported in SI-K 1.0 and SI K 1.1 (since XD 1.1 will seemingly need manual acks and we don't want to force a SI-K upgrade until 1.2)

mbogoevici:
Closed in favour of https://github.com/spring-projects/spring-integration-kafka/pull/21

Issue #19, opened 2015-02-25
Version 1.0.0.RC1 does not work with Kafka 0.8.2 on consumer side


mohitarora:
I have 2 projects 

https://github.com/mohitarora/es-cqrs-example/tree/master/cart-command-services - This one produces messages and publishes to Kafka and works perfectly fine. 

https://github.com/mohitarora/es-cqrs-example/tree/master/cart-query-services - This one consumers messages published by previous project.

Both these projects uses Spring Boot. There is no error but the messages are never consumed. I also don't see any error but I am sure beans are not initialized.

Am i doing something wrong?

artembilan:
@mohitarora , first of all thank you for paying attention to this project!

Let me ask for more info from you:
1. Let us know how it works with 1.0.0.RELEASE
2. Why do you make such a conclusion: " but I am sure beans are not initialized." ? Is there anything which does not say that it has been started or anything similar? Can't you access to some beans from your app?
3. Have you tried to send messages to the Kafka topic directly without that publisher application ?

Note: we have plans to upgrade to the Kafka `0.8.2` including new features, but it will be for our `1.1.0` version. So, to be honest we haven't tested our solution against Kafka 0.8.2 :smile:

mohitarora:
I upgraded to 1.0.0.RELEASE, no luck

Producer is still working fine (https://github.com/mohitarora/es-cqrs-example/blob/master/cart-command-services/src/main/resources/kafka.xml)

Consumer is not (https://github.com/mohitarora/es-cqrs-example/blob/master/cart-query-services/src/main/resources/kafka.xml)

My guess is I am doing something stupid on consumer side, Can you please check my kafka.xml on consumer side.

I tried sending messages to Kafka topic directly without producer application, still no luck.

Why i say the beans are not initialized. I have a break point in line 58 of KafkaConsumerContext.java and when i run in debug mode, control never reaches there. I must re-word, beans are getting initialized else application will not start but they are not getting correctly initialized. I don't see any kafka related stuff on spring boot application startup logs.

artembilan:
@mohitarora , can you confirm that you app works well with `Kafka-0.8.1.1`?
And we go ahead to seek an issue with 0.8.2 compatibility.
Thank you very much!

mohitarora:
This issue is resolved with 1.1.1.RELEASE

PR #21, opened 2015-03-05
INTEXT-143 Add support for manual acknowledgment of messages

mbogoevici:
- Introduce an Acknowledgment object that message processors can invoke;
- Introduce an AcknowedgingMessageListener variation of the MessageListener that receives an Acknowledgment reference for the processed message;
- Add 'autoCommitOffset' settings to the KafkaMessageListenerContainer and KafkaMessageDrivenChannelAdapter, and the ability to inject a MessageListener or an AcknowledgingMessageListener in either (the allowed type depending on the offset management strategy)
- Prepopulate a message header for SI messages created by the KafkaMessageDrivenChannelAdapter if autoCommit is enabled or disabled;
- OffsetManager only sets values that are higher than the ones already set in a session (barring reset) - this is to prevent asynchrous acks to mistakenly revert checkpoints

### DIFF
--- a/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
+++ b/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
@@ -65,7 +65,6 @@ public class KafkaMessageDrivenChannelAdapter extends MessageProducerSupport imp
 
 	@Override
 	protected void onInit() {
-		this.messageListenerContainer.setAutoCommitOffset(autoCommitOffset);
 		this.messageListenerContainer.setMessageListener(autoCommitOffset ?
 				new AutoAcknowledgingChannelForwardingMessageListener()
 				: new AcknowledgingChannelForwardingMessageListener());
--- a/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
+++ b/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
@@ -22,11 +22,13 @@ import kafka.serializer.DefaultDecoder;
 import org.springframework.integration.context.OrderlyShutdownCapable;
 import org.springframework.integration.endpoint.MessageProducerSupport;
 import org.springframework.integration.kafka.core.KafkaMessageMetadata;
+import org.springframework.integration.kafka.listener.AbstractDecodingAcknowledgingMessageListener;
 import org.springframework.integration.kafka.listener.AbstractDecodingMessageListener;
+import org.springframework.integration.kafka.listener.Acknowledgment;
 import org.springframework.integration.kafka.listener.KafkaMessageListenerContainer;
 import org.springframework.integration.kafka.support.KafkaHeaders;
+import org.springframework.integration.support.AbstractIntegrationMessageBuilder;
 import org.springframework.messaging.Message;
-import org.springframework.messaging.support.MessageBuilder;
 import org.springframework.util.Assert;
 
 /**
@@ -39,6 +41,8 @@ public class KafkaMessageDrivenChannelAdapter extends MessageProducerSupport imp
 	private Decoder<?> keyDecoder = new DefaultDecoder(null);
 
 	private Decoder<?> payloadDecoder = new DefaultDecoder(null);
+
+	private boolean autoCommitOffset = true;
 
 	public KafkaMessageDrivenChannelAdapter(KafkaMessageListenerContainer messageListenerContainer) {
 		Assert.notNull(messageListenerContainer);
@@ -55,9 +59,16 @@ public class KafkaMessageDrivenChannelAdapter extends MessageProducerSupport imp
 		this.payloadDecoder = payloadDecoder;
 	}
 
+	public void setAutoCommitOffset(boolean autoCommitOffset) {
+		this.autoCommitOffset = autoCommitOffset;
+	}
+
 	@Override
 	protected void onInit() {
-		this.messageListenerContainer.setMessageListener(new ChannelForwardingMessageListener());
+		this.messageListenerContainer.setAutoCommitOffset(autoCommitOffset);
+		this.messageListenerContainer.setMessageListener(autoCommitOffset ?
+				new AutoAcknowledgingChannelForwardingMessageListener()
+				: new AcknowledgingChannelForwardingMessageListener());
 		super.onInit();
 	}
 
@@ -88,25 +99,47 @@ public class KafkaMessageDrivenChannelAdapter extends MessageProducerSupport imp
 	}
 
 	@SuppressWarnings("rawtypes")
-	private class ChannelForwardingMessageListener extends AbstractDecodingMessageListener {
+	private class AutoAcknowledgingChannelForwardingMessageListener extends AbstractDecodingMessageListener {
 
 		@SuppressWarnings("unchecked")
-		public ChannelForwardingMessageListener() {
+		public AutoAcknowledgingChannelForwardingMessageListener() {
 			super(keyDecoder, payloadDecoder);
 		}
 
 		@Override
 		public void doOnMessage(Object key, Object payload, KafkaMessageMetadata metadata) {
-			Message<Object> message = getMessageBuilderFactory()
-					.withPayload(payload)
-					.setHeader(KafkaHeaders.MESSAGE_KEY, key)
-					.setHeader(KafkaHeaders.TOPIC, metadata.getPartition().getTopic())
-					.setHeader(KafkaHeaders.PARTITION_ID, metadata.getPartition().getId())
-					.setHeader(KafkaHeaders.OFFSET, metadata.getOffset())
-					.build();
-			KafkaMessageDrivenChannelAdapter.this.sendMessage(message);
+			KafkaMessageDrivenChannelAdapter.this.sendMessage(toMessage(key, payload, metadata, null));
 		}
 
+	}
+
+	@SuppressWarnings("rawtypes")
+	private class AcknowledgingChannelForwardingMessageListener extends AbstractDecodingAcknowledgingMessageListener {
+
+		@SuppressWarnings("unchecked")
+		public AcknowledgingChannelForwardingMessageListener() {
+			super(keyDecoder, payloadDecoder);
+		}
+
+		@Override
+		public void doOnMessage(Object key, Object payload, KafkaMessageMetadata metadata, Acknowledgment acknowledgment) {
+			KafkaMessageDrivenChannelAdapter.this.sendMessage(toMessage(key, payload, metadata, acknowledgment));
+		}
+
+	}
+
+	private Message<Object> toMessage(Object key, Object payload, KafkaMessageMetadata metadata, Acknowledgment acknowledgment) {
+		AbstractIntegrationMessageBuilder<Object> messageBuilder = getMessageBuilderFactory()
+				.withPayload(payload)
+				.setHeader(KafkaHeaders.MESSAGE_KEY, key)
+				.setHeader(KafkaHeaders.TOPIC, metadata.getPartition().getTopic())
+				.setHeader(KafkaHeaders.PARTITION_ID, metadata.getPartition().getId())
+				.setHeader(KafkaHeaders.OFFSET, metadata.getOffset())
+				.setHeader(KafkaHeaders.NEXT_OFFSET, metadata.getNextOffset());
+		if (acknowledgment != null) {
+			messageBuilder.setHeader(KafkaHeaders.ACKNOWLEDGMENT, acknowledgment);
+		}
+		return messageBuilder.build();
 	}
 
 }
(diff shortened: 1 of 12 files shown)


src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
```diff
@@ -55,9 +59,16 @@ public void setPayloadDecoder(Decoder<?> payloadDecoder) {
...
 	}
 
+	public void setAutoCommitOffset(boolean autoCommitOffset) {
+		this.autoCommitOffset = autoCommitOffset;
+	}
+
 	@Override
 	protected void onInit() {
-		this.messageListenerContainer.setMessageListener(new ChannelForwardingMessageListener());
+		this.messageListenerContainer.setAutoCommitOffset(autoCommitOffset);
```
artembilan: Can't we make `KafkaMessageListenerContainer` as smart for that `autoCommitOffset` flag?
I mean determine it by the type of provided `messageListener` instead of the fact to say that end-user is fool :smile: .
And avoid that hard work for the `Assert.isTrue()` in those listenerContainer participants.

mbogoevici: That's a good idea. I had the flag initially and the check later, but they can be collapsed into one. That'd be much cleaner for sure.

src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -142,12 +142,21 @@ public void setOffsetManager(OffsetManager offsetManager) {
...
 
-	public MessageListener getMessageListener() {
+	public Object getMessageListener() {
 		return messageListener;
 	}
 
-	public void setMessageListener(MessageListener messageListener) {
-		this.messageListener = messageListener;
+	public void setMessageListener(Object messageListener) {
+		if (autoCommitOffset) {
```
artembilan: Having this check in setter doesn't guarantee that `setAutoCommitOffset` will be invoked latter.
That's why I suggest to get rid of it at all.

mbogoevici: Yeah, let's go that route.

src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
mbogoevici: (And I agree about this being wrong, as there's no telling in which order the two set methods are called. So I'm onboard on getting rid of the autoCommitOffset setting. In fact, we can just push that check down to the `QueueingMessageListenerInvoker`.

src/main/java/org/springframework/integration/kafka/support/KafkaHeaders.java
```diff
@@ -32,4 +33,8 @@
 
 	public static final String OFFSET = PREFIX + "_offset";
 
+	public static final String NEXT_OFFSET = PREFIX + "nextOffset";
+
+	public static final String ACKNOWLEDGMENT = PREFIX + "acknowledgment";
```
artembilan: I don't see fixes for the extra `_`, so this PR is for the `1.0.x` and we start `1.1.x` afterward.
Correct ?

src/main/java/org/springframework/integration/kafka/listener/AbstractOffsetManager.java
```diff
@@ -54,6 +54,8 @@
 
 	protected Map<Partition, Long> initialOffsets;
 
+	protected Map<Partition, Long> highestUpdatedOffsets = new ConcurrentHashMap<Partition, Long>();
```
artembilan: Can't we collapse this `*Offsets` Maps to the single one?
I don't `read` usage of the `highestUpdatedOffsets` hence I can simply update that `initialOffsets` for the new `highestUpdatedOffset` value.
WDYT ?

src/main/java/org/springframework/integration/kafka/listener/DefaultAcknowledgment.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.integration.kafka.core.KafkaMessage;
+import org.springframework.integration.kafka.core.Partition;
+
+/**
+ * Default implementation for an {@link Acknowledgment} that defers to an underlying
+ * {@link OffsetManager}.
+ *
+ * @author Marius Bogoevici
+ */
+public class DefaultAcknowledgment implements Acknowledgment {
```
artembilan: That's excellent for the `late updateOffset` ! :+1:

src/test/java/org/springframework/integration/kafka/listener/KafkaMessageDrivenChannelAdapterTests.java
```diff
@@ -124,4 +130,92 @@ public boolean send(Message<?> message, long timeout) {
...
+		}
+
+		// now we did acknowledge them in the reverse order. This way we check that only the highest value was acknowledged
+		for (Acknowledgment acknowledgment : FastList.newList(acknowledgments).reverseThis()) {
+			acknowledgment.acknowledge();
+		}
+
+		// now they are all acknowledged
+		for (Partition readPartition : readPartitions) {
+			assertThat(metadataStore.get(offsetManager.generateKey(readPartition)), equalTo(String.valueOf(20)));
```
artembilan: Would you mind explaining why this `lastOffest` is `20` ?

src/test/java/org/springframework/integration/kafka/listener/SingleBrokerWithManualAckTests.java
```diff
@@ -0,0 +1,171 @@
...
+		kafkaMessageListenerContainer.setMaxFetch(100);
+		kafkaMessageListenerContainer.setConcurrency(2);
+
+		kafkaMessageListenerContainer.setMessageListener(new AcknowledgingMessageListener() {
+			@Override
+			public void onMessage(KafkaMessage message, Acknowledgment acknowledgment) {
+				// do nothing
+			}
+		});
+	}
```
artembilan: So, according to our discussion above these `expected = IllegalArgumentException.class` tests are redundant :smile:

src/test/java/org/springframework/integration/kafka/listener/SingleBrokerWithManualAckTests.java
```diff
@@ -0,0 +1,171 @@
...
+
+		kafkaMessageListenerContainer.start();
+
+		createStringProducer(0).send(createMessages(100, TEST_TOPIC));
+
+		latch.await((expectedMessageCount/5000) + 1, TimeUnit.MINUTES);
+		kafkaMessageListenerContainer.stop();
+
+		assertThat(receivedData.valuesView().toList(), hasSize(expectedMessageCount));
+		assertThat(latch.getCount(), equalTo(0L));
```
artembilan: One more time:
```
assertTrue(latch.await)
```
is enough to be sure that latch has been counted down

src/test/java/org/springframework/integration/kafka/listener/SingleBrokerWithManualAckTests.java
```diff
@@ -0,0 +1,171 @@
...
+
+		validateMessageReceipt(receivedData, 2, 5, 100, expectedMessageCount, readPartitions, 1);
+
+		// at this point, all messages have been processed but not acknowledged
+		for (Partition readPartition : readPartitions) {
+			assertThat(metadataStore.get(offsetManager.generateKey(readPartition)), nullValue());
+		}
+
+		// now we did acknowledge them in the reverse order. This way we check that only the highest value was acknowledged
+		for (Acknowledgment acknowledgment : FastList.newList(acknowledgments).reverseThis()) {
```
artembilan: What is the reason to `reverse` ?

artembilan:
That's all for now. Will take a look one more time after the fixing commit.
However I must confirm that it is great stuff and provides some concepts to be used in other places.

Thank you very much for sharing your wisdom!

src/test/java/org/springframework/integration/kafka/listener/KafkaMessageDrivenChannelAdapterTests.java
mbogoevici: We do not acknowledge to the OffsetManager the last consumed offset, we acknowledge the next offset - so that a new consumer should start where the other left off. If we didn't do that, then we would have to replay the last message in order to see where the next consumer should start.

mbogoevici: So the next offset after processing 20 messages (100 divided by 5, the number of partitions) is 20.

src/test/java/org/springframework/integration/kafka/listener/SingleBrokerWithManualAckTests.java
mbogoevici: Normally, offsets would be processed incrementally, but by reversing them during the test we validate that older offsets don't overwrite newer offsets too.

mbogoevici:
Thanks for the review, @artembilan - I'll push some changes today and perhaps we can revise on Monday.

src/main/java/org/springframework/integration/kafka/support/KafkaHeaders.java
mbogoevici: Yes, we'll make that a separate fix so that we can have a 1.0.x with support for this.

src/main/java/org/springframework/integration/kafka/listener/AbstractDecodingAcknowledgingMessageListener.java
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.integration.kafka.core.KafkaMessageMetadata;
+import org.springframework.integration.kafka.util.MessageUtils;
+
+/**
+ * Base {@link AcknowledgingMessageListener} implementation that decodes the key and the payload using the supplied {@link Decoder}s.
+ *
+ * Users of this class must extend it and implement {@code doOnMessage} and must supply {@link Decoder}
+ * implementations for both the key and the payload.
+ *
+ * @author Marius Bogoevici
```
artembilan: `@Since 1.0.1` ?

src/main/java/org/springframework/integration/kafka/listener/Acknowledgment.java
```diff
@@ -0,0 +1,20 @@
+package org.springframework.integration.kafka.listener;
```
artembilan: Need Copyright

artembilan:
After the merge of this PR we need initiate a new master `1.1.x` and PR the fix for XSD (and others, if you want). Plus we shouldn't forget to update Docs.

And stat to see what's going on with Kafka-0.8.2 - https://github.com/spring-projects/spring-integration-kafka/issues/19

mbogoevici:
Let's make that a separate issue. It could be that 0.8.2 is not backwards compatible for the high level API.

mbogoevici:
I have a few more things that would like to fix before moving to 1.1

src/main/java/org/springframework/integration/kafka/listener/AbstractOffsetManager.java
mbogoevici: Well, I'd honestly wouldn't want to mix things, and keep the two separate. It could be wasteful, but i'd avoid the interference, to be honest.

artembilan: OK. Let it be!

src/main/java/org/springframework/integration/kafka/listener/AbstractDecodingAcknowledgingMessageListener.java
artembilan: Isn't addressed

src/main/java/org/springframework/integration/kafka/listener/DefaultAcknowledgment.java
```diff
@@ -0,0 +1,51 @@
...
+package org.springframework.integration.kafka.listener;
+
+import org.springframework.integration.kafka.core.KafkaMessage;
+import org.springframework.integration.kafka.core.Partition;
+
+/**
+ * Default implementation for an {@link Acknowledgment} that defers to an underlying
+ * {@link OffsetManager}.
+ *
+ * @author Marius Bogoevici
```
artembilan: `@since 1.0.1`

src/test/java/org/springframework/integration/kafka/listener/SingleBrokerWithManualAckTests.java
artembilan: Isn't addressed

artembilan:
That's all from my perspective.
All those minor concerns can be addressed during merge.
Plus code style for the `120` line length :smile:

src/test/java/org/springframework/integration/kafka/listener/SingleBrokerWithManualAckTests.java
mbogoevici: I'd prefer not to address that now. It's not fundamentally incorrect and requires changing all the tests to match the same pattern, but, the actual problem is that if the latch doesn't actually count down to zero, it would throw an exception and prevent the kafkaMessageListenerContainer to stop, which can cause problems for other tests. Until the tests are reworked, I'd prefer to leave it as is.

garyrussell:
Merged as 08bbe25a8815bceee3270bc17744ace6e56ccdc4

PR #24, opened 2015-03-18
[1.0.1][1.1]INTEXT-153 Properly stop the KafkaMessageListenerContainer

mbogoevici:
- ensure that the fetch loop is exited immediately after the listener container is stopped;
- do not dispatch messages for processing once components are stopped;


artembilan:
Nothing to complain about.
Merged and chery-picked after rebase.

PR #25, opened 2015-03-18
[1.1] INTEXT-152 Improve message generation strategy

mbogoevici:
- upgrade to Spring Integration 4.1 and Spring 4.1 for MessageHeaders subclassing support;
- by default, the KafkaMessageDrivenChannelAdapter will create an optimized version of KafkaMessage;
- add settings for avoiding expensive ID and timestamp generation by default, with the option of turning them on;
- add a setting for reverting to superclass MessageBuilderFactory where that is necessary;

### DIFF
--- a/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
+++ b/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
@@ -16,6 +16,14 @@
 
 package org.springframework.integration.kafka.inbound;
 
+import java.util.Map;
+import java.util.UUID;
+
+import com.gs.collections.api.map.MutableMap;
+import com.gs.collections.impl.map.mutable.UnifiedMap;
+import kafka.serializer.Decoder;
+import kafka.serializer.DefaultDecoder;
+
 import org.springframework.integration.context.OrderlyShutdownCapable;
 import org.springframework.integration.endpoint.MessageProducerSupport;
 import org.springframework.integration.kafka.core.KafkaMessageMetadata;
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
```diff
@@ -63,11 +77,49 @@ public void setAutoCommitOffset(boolean autoCommitOffset) {
...
+	 * @param generateTimestamp true if a timestamp should be generated
+	 */
+	public void setGenerateTimestamp(boolean generateTimestamp) {
+		this.generateTimestamp = generateTimestamp;
+	}
+
+	/**
+	 * Use the {@link MessageBuilderFactory} returned by {@link #getMessageBuilderFactory()} to create messages.
+	 *
+	 * @param useMessageBuilderFactory true if the @link MessageBuilderFactory} returned by
```
artembilan: Missed `{` before the `@link`.
Minor and can be addressed during merge with other simple polishing, like "Please" in JavaDocs, or `we use this for` below :smile: .

But..
Shouldn't we expose these new options to the Namespace support?

mbogoevici: I think they should be, can we add a 1.1 enhancement for later ?

mbogoevici: at least a separate JIRA

src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
```diff
@@ -128,18 +180,76 @@ public void doOnMessage(Object key, Object payload, KafkaMessageMetadata metadat
...
+		}
+
+		@Override
+		public MessageHeaders getHeaders() {
+			return messageHeaders;
+		}
+
+	}
+
+	private class KafkaMessageHeaders extends MessageHeaders {
```
artembilan: `serialVersionUID` is required. Or at least `@SuppressWarnings("serial")`

src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
artembilan: OK. No problem: https://jira.spring.io/browse/INTEXT-154 !
I'm merging it with those polishing

artembilan:
Merged as https://github.com/spring-projects/spring-integration-kafka/commit/0182247f63e0eaf40a30798be372ac49ae40ac9b

PR #27, opened 2015-03-21
Improve performance of header generation

mbogoevici:
Using a regular HashMap instead of GS Collections' regular map improves performance, as the latter is more expensive to create.

### DIFF
--- a/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
+++ b/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
@@ -16,7 +16,11 @@
 
 package org.springframework.integration.kafka.inbound;
 
+import java.util.HashMap;
 import java.util.Map;
+
+import kafka.serializer.Decoder;
+import kafka.serializer.DefaultDecoder;
 
 import org.springframework.integration.context.OrderlyShutdownCapable;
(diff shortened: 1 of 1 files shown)


src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
```diff
@@ -185,21 +190,22 @@ public void doOnMessage(Object key, Object payload, KafkaMessageMetadata metadat
...
+		final Map<String, Object> headers = new HashMap<String, Object>();
+
+		headers.put(KafkaHeaders.MESSAGE_KEY, key);
+		headers.put(KafkaHeaders.TOPIC, metadata.getPartition().getTopic());
+		headers.put(KafkaHeaders.PARTITION_ID, metadata.getPartition().getId());
+		headers.put(KafkaHeaders.OFFSET, metadata.getOffset());
+		headers.put(KafkaHeaders.NEXT_OFFSET, metadata.getNextOffset());
 
 		// pre-set the message id header if set to not generate
-		if (!this.generateMessageId) {
```
artembilan: Why don't you like `this.` :smile:  ?

mbogoevici: :) I can put it back.

artembilan:
Merged after squesh and rebase

PR #28, opened 2015-03-25
INTEXT 156-157

mbogoevici:
- `KafkaMessageListenerContainer#stop()` returns only after the container has been effectively stopped, i.e. no more messages are dispatched to the listeners;
- add a configurable timeout property to prevent the previous operation from blocking indefinitely;
- create separate XSDs for 1.0 and 1.1 schemas;


src/main/resources/org/springframework/integration/config/xml/spring-integration-kafka-1.1.xsd
```diff
@@ -0,0 +1,708 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
artembilan: No reason in the separate file: just rename to the new version is enough.
See how we live with that in the SI Core.

src/main/java/org/springframework/integration/kafka/listener/QueueingMessageListenerInvoker.java
```diff
@@ -28,7 +32,9 @@
  *
  * @author Marius Bogoevici
  */
-class QueueingMessageListenerInvoker implements Runnable, Lifecycle {
+class QueueingMessageListenerInvoker implements Runnable {
+
+	private static Logger log = LogManager.getLogger(QueueingMessageListenerInvoker.class);
```
artembilan: This is redundant.
Otherwise everything looks good.
Not sure why we have missed that. Similar `stop` issue we faced with `RedisQueueMessageDrivenEndpoint` some days ago... And thanks to XD also.

src/test/java/org/springframework/integration/kafka/listener/KafkaMessageDrivenChannelAdapterTests.java
```diff
@@ -220,4 +222,131 @@ public boolean send(Message<?> message, long timeout) {
...
+
+		kafkaMessageDrivenChannelAdapter.setAutoCommitOffset(false);
+		kafkaMessageDrivenChannelAdapter.afterPropertiesSet();
+		kafkaMessageDrivenChannelAdapter.start();
+
+		createStringProducer(NoCompressionCodec$.MODULE$.codec()).send(createMessages(1, TEST_TOPIC));
+
+		long startTime = System.currentTimeMillis();
+		kafkaMessageDrivenChannelAdapter.stop();
+		Assert.assertThat(System.currentTimeMillis() - startTime, Matchers.greaterThan(1500L));
```
artembilan: > Expected: a value greater than <1500L>
     but: <1L> was less than <1500L>

These new tests are very sensitive to time and enough long (`Thread.sleep(2000L)`).

Need to come up with some `barrier` variable that the `stop()` is really finishing only after `handleMessage`

src/main/java/org/springframework/integration/kafka/listener/QueueingMessageListenerInvoker.java
```diff
@@ -90,19 +98,19 @@ public void enqueue(KafkaMessage message) {
...
-	@Override
 	public void start() {
 		this.running = true;
 	}
 
-	@Override
-	public void stop() {
+	public void stop(long stopTimeout) {
+		shutdownLatch = new CountDownLatch(1);
 		this.running = false;
```
artembilan: Shouldn't this code be moved after that `shutdownLatch.await`? To allow to process all those in-flight `messages` ?

src/test/java/org/springframework/integration/kafka/listener/KafkaMessageDrivenChannelAdapterTests.java
```diff
@@ -220,4 +222,131 @@ public boolean send(Message<?> message, long timeout) {
...
+		kafkaMessageListenerContainer.setConcurrency(1);
+		kafkaMessageListenerContainer.setStopTimeout(3000);
+
+		int expectedMessageCount = 1;
+
+
+		KafkaMessageDrivenChannelAdapter kafkaMessageDrivenChannelAdapter =
+				new KafkaMessageDrivenChannelAdapter(kafkaMessageListenerContainer);
+
+		final CountDownLatch latch = new CountDownLatch(expectedMessageCount);
```
artembilan: This `latch` isn't covered in test.

artembilan:
That's all from my side

src/main/java/org/springframework/integration/kafka/listener/QueueingMessageListenerInvoker.java
mbogoevici: I don't think we should. Processing the messages can take fairly long and we'd want to return as soon as possible. Given that no offsets get updated as a result, we can replay the.

src/test/java/org/springframework/integration/kafka/listener/KafkaMessageDrivenChannelAdapterTests.java
mbogoevici: Actually, that's not correct. Stop() may finish before handleMessage() if it times out.

src/main/resources/org/springframework/integration/config/xml/spring-integration-kafka-1.1.xsd
mbogoevici: Ok - Spring Core follows a different policy. I'll remove the 1.0 file then.

src/main/java/org/springframework/integration/kafka/listener/QueueingMessageListenerInvoker.java
artembilan: Yes, I see by your code.
What is the benefit of the change afterward?
Consider we use hands off `MessageChannel`.
Maybe an issue with the adapter's `phase` ?

Would you mind sharing the StackTrace in the JIRA?

src/test/java/org/springframework/integration/kafka/listener/KafkaMessageDrivenChannelAdapterTests.java
artembilan: Right, I understand.
But it is really difficult to come up with a good timing test-case...

src/test/java/org/springframework/integration/kafka/listener/KafkaMessageDrivenChannelAdapterTests.java
mbogoevici: I'll remove it.

src/main/java/org/springframework/integration/kafka/listener/QueueingMessageListenerInvoker.java
mbogoevici: This is in a scenario that doesn't use Application contexts. So, stop() shouldn't return while there are messages in flight. That's just wrong.

mbogoevici: We can add an option to drain the internal before stopping but I don't think that should be the default.

mbogoevici:
Updated.

src/main/java/org/springframework/integration/kafka/listener/QueueingMessageListenerInvoker.java
```diff
@@ -90,19 +96,19 @@ public void enqueue(KafkaMessage message) {
...
-	}
-
-	@Override
-	public boolean isRunning() {
-		return this.running;
+		try {
+			shutdownLatch.await(stopTimeout, TimeUnit.MILLISECONDS);
+		}
+		catch (InterruptedException e) {
+			e.printStackTrace();
```
artembilan:
```
Thread.currentThread().interrupt();
```
Instead.

mbogoevici: Right.

artembilan:
Merged as https://github.com/spring-projects/spring-integration-kafka/commit/c27bdb1c521206ca2c4f942757e913fb251c30dc

PR #31, opened 2015-04-15
Change expected type of listener-container attribute for message-drive-channel-adapter

moonkev:
Currently the expected type of the listener-container attribute of the message-driven-channel-adapter for the kafka-int schema is org.springframework.integration.kafka.core.ConnectionFactory instead of the true required type of org.springframework.integration.kafka.listener.KafkaMessageListenerContainer.  While not enforced at run time, this does report as a break in spring schema parsers, such as the one included with Intellij IDEA.

Regards,
Kevin

### DIFF
--- a/src/main/resources/org/springframework/integration/config/xml/spring-integration-kafka-1.1.xsd
+++ b/src/main/resources/org/springframework/integration/config/xml/spring-integration-kafka-1.1.xsd
@@ -551,7 +551,7 @@
 					</xsd:documentation>
 					<xsd:appinfo>
 						<tool:annotation kind="ref">
-							<tool:expected-type type="org.springframework.integration.kafka.core.ConnectionFactory"/>
+							<tool:expected-type type="org.springframework.integration.kafka.listener.KafkaMessageListenerContainer"/>
 						</tool:annotation>
 					</xsd:appinfo>
 				</xsd:annotation>


artembilan:
Thank you very much for the feedback and PR, too, @moonkev , but before review your (even such a simple) work, we should be sure that you have signed CLA: https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.md

moonkev:
I have signed and agree to the terms of the SpringSource Individual
Contributor License Agreement.

artembilan:
@moonkev , thank you for the confirmation.
Even if the fix is so simple and correct, we can't merge your commit because there is need separate branch, e.g. `fixXSDType` not `master`.
Please, refer to that Contribution Guideline one more time.
Of course, it is better to have a JIRA or GH issue for that, but I don't see reason for such a simple fix.
So, just create a new branch based on `master` and make a new PR.
This one must be closed after that, because it isn't valid by merging process rules.

garyrussell:
Actually for such a trivial fix, there is no need to go through all that, but in future, please use the guidelines.

Thanks for your contribution, we look forward to more!

artembilan:
Yes, @garyrussell is right. So, nevermind and we go ahead to merge your fix.
Thank you one more time and don't hesitate to contribute more!

artembilan:
Merged as https://github.com/spring-projects/spring-integration-kafka/commit/aa753a372a064add7ab07dcc9359510a01baefcb

moonkev:
Thanks gents for the merge on this push request.  I will make sure to follow the guidelines for any future contributions.

PR #34, opened 2015-05-13
INTEXT-168 Upgrade to Kafka 0.8.2.1

mbogoevici:
- Replace producers with the new producer API
    - Change KafkaProducerContext and ProducerConfiguration to match the new Producer's behaviour (i.e. providing partition as argument)
    - support using the `partitionId` message header for targetting specific partitions;
    - allow using a SpEL expression for partitioning;
    - remove configurations that do not apply anymore, i.e. async (new producer is always async), batch size as message count, etc.
    - add new configuration options wherever necessary;
    - tweak schema - allow reuse of Encoders and Partitioners if available;

    TODO: Add missing unit tests;

### DIFF
--- a/src/main/java/org/springframework/integration/kafka/config/xml/KafkaProducerContextParser.java
+++ b/src/main/java/org/springframework/integration/kafka/config/xml/KafkaProducerContextParser.java
@@ -18,6 +18,8 @@ package org.springframework.integration.kafka.config.xml;
 
 import java.util.Map;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.w3c.dom.Element;
 
 import org.springframework.beans.BeanMetadataElement;
@@ -31,6 +33,8 @@ import org.springframework.integration.kafka.support.KafkaProducerContext;
 import org.springframework.integration.kafka.support.ProducerConfiguration;
 import org.springframework.integration.kafka.support.ProducerFactoryBean;
 import org.springframework.integration.kafka.support.ProducerMetadata;
+import org.springframework.integration.kafka.util.EncoderAdaptingSerializer;
+import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 import org.springframework.util.xml.DomUtils;
 
@@ -41,6 +45,8 @@ import org.springframework.util.xml.DomUtils;
  * @since 0.5
  */
 public class KafkaProducerContextParser extends AbstractSimpleBeanDefinitionParser {
+
+	private static final Log log = LogFactory.getLog(KafkaProducerContextParser.class);
 
 	@Override
 	protected Class<?> getBeanClass(final Element element) {
@@ -65,7 +71,7 @@ public class KafkaProducerContextParser extends AbstractSimpleBeanDefinitionPars
 	}
 
 	private void parseProducerConfigurations(Element topics, ParserContext parserContext,
-			BeanDefinitionBuilder builder, Element parentElem) {
+																					 BeanDefinitionBuilder builder, Element parentElem) {
 		Map<String, BeanMetadataElement> producerConfigurationsMap = new ManagedMap<String, BeanMetadataElement>();
 
 		for (Element producerConfiguration : DomUtils.getChildElementsByTagName(topics, "producer-configuration")) {
@@ -73,22 +79,52 @@ public class KafkaProducerContextParser extends AbstractSimpleBeanDefinitionPars
 			BeanDefinitionBuilder producerMetadataBuilder =
 					BeanDefinitionBuilder.genericBeanDefinition(ProducerMetadata.class);
 			producerMetadataBuilder.addConstructorArgValue(producerConfiguration.getAttribute("topic"));
-			IntegrationNamespaceUtils.setReferenceIfAttributeDefined(producerMetadataBuilder, producerConfiguration,
-					"value-encoder");
-			IntegrationNamespaceUtils.setReferenceIfAttributeDefined(producerMetadataBuilder, producerConfiguration,
-					"key-encoder");
-			IntegrationNamespaceUtils.setValueIfAttributeDefined(producerMetadataBuilder, producerConfiguration,
-					"key-class-type");
-			IntegrationNamespaceUtils.setValueIfAttributeDefined(producerMetadataBuilder, producerConfiguration,
-					"value-class-type");
+			producerMetadataBuilder.addConstructorArgValue(producerConfiguration.getAttribute("key-class-type"));
+			producerMetadataBuilder.addConstructorArgValue(producerConfiguration.getAttribute("value-class-type"));
+
+			String keySerializer = producerConfiguration.getAttribute("key-serializer");
+			String keyEncoder = producerConfiguration.getAttribute("key-encoder");
+			Assert.isTrue((StringUtils.hasText(keySerializer) ^ StringUtils.hasText(keyEncoder)),
+					"Exactly one of 'key-serializer' or 'key-encoder' can be specified");
+			if (StringUtils.hasText(keyEncoder)) {
+				if (log.isWarnEnabled()) {
+					log.warn("'key-encoder' is a deprecated option, use 'key-serializer' instead.");
+				}
+				BeanDefinitionBuilder encoderAdaptingSerializerBean = BeanDefinitionBuilder.genericBeanDefinition(EncoderAdaptingSerializer.class);
+				encoderAdaptingSerializerBean.addConstructorArgReference(keyEncoder);
+				producerMetadataBuilder.addConstructorArgValue(encoderAdaptingSerializerBean.getBeanDefinition());
+			}
+			else {
+				producerMetadataBuilder.addConstructorArgReference(keySerializer);
+			}
+
+			String valueSerializer = producerConfiguration.getAttribute("value-serializer");
+			String valueEncoder = producerConfiguration.getAttribute("value-encoder");
+			Assert.isTrue((StringUtils.hasText(valueSerializer) ^ StringUtils.hasText(valueEncoder)),
+					"Exactly one of 'value-serializer' or 'value-encoder' must be specified");
+			if (StringUtils.hasText(valueEncoder)) {
+				if (log.isWarnEnabled()) {
+					log.warn("'value-encoder' is a deprecated option, use 'value-serializer' instead.");
+				}
+				BeanDefinitionBuilder encoderAdaptingSerializerBean = BeanDefinitionBuilder.genericBeanDefinition(EncoderAdaptingSerializer.class);
+				encoderAdaptingSerializerBean.addConstructorArgReference(valueEncoder);
+				producerMetadataBuilder.addConstructorArgValue(encoderAdaptingSerializerBean.getBeanDefinition());
+			}
+			else {
+				producerMetadataBuilder.addConstructorArgReference(valueSerializer);
+			}
+
 			IntegrationNamespaceUtils.setReferenceIfAttributeDefined(producerMetadataBuilder, producerConfiguration,
 					"partitioner");
+			if (StringUtils.hasText(producerConfiguration.getAttribute("partitioner"))) {
+				if (log.isWarnEnabled()) {
+					log.warn("'partitioner' is a deprecated option. Use the 'kafka_partitionId' message header or the partition argument in the send() or convertAndSend() methods");
+				}
+			}
 			IntegrationNamespaceUtils.setValueIfAttributeDefined(producerMetadataBuilder, producerConfiguration,
-					"compression-codec");
+					"compression-type");
 			IntegrationNamespaceUtils.setValueIfAttributeDefined(producerMetadataBuilder, producerConfiguration,
-					"async");
-			IntegrationNamespaceUtils.setValueIfAttributeDefined(producerMetadataBuilder, producerConfiguration,
-					"batch-num-messages");
+					"batch-bytes");
 			AbstractBeanDefinition producerMetadataBeanDefinition = producerMetadataBuilder.getBeanDefinition();
 
 			String producerPropertiesBean = parentElem.getAttribute("producer-properties");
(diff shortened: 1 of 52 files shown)


src/main/java/org/springframework/integration/kafka/listener/KafkaTopicOffsetManager.java
```diff
@@ -385,28 +383,10 @@ private void checkAndAddData(KafkaMessage kafkaMessage) {
...
-			producerMetadata.setCompressionCodec(this.compressionCodec);
-			additionalProps.put("request.required.acks", String.valueOf(this.requiredAcks));
-			additionalProps.put("queue.buffering.max.ms", String.valueOf(this.maxQueueBufferingTime));
-		}
-
-		ProducerFactoryBean<Key, Long> producerFB
-				= new ProducerFactoryBean<Key, Long>(producerMetadata, leader.toString(), additionalProps);
-
-		this.producer = producerFB.getObject();
+		additionalProps.setProperty("bootstrap.servers", leader.toString());
```
mbogoevici: Should use constant.

src/main/resources/META-INF/spring.schemas
```diff
@@ -1,2 +1,2 @@
-http\://www.springframework.org/schema/integration/kafka/spring-integration-kafka-1.1.xsd=org/springframework/integration/config/xml/spring-integration-kafka-1.1.xsd
-http\://www.springframework.org/schema/integration/kafka/spring-integration-kafka.xsd=org/springframework/integration/config/xml/spring-integration-kafka-1.1.xsd
+http\://www.springframework.org/schema/integration/kafka/spring-integration-kafka-1.1.xsd=org/springframework/integration/config/xml/spring-integration-kafka-1.2.xsd
```
mbogoevici: Should be 1.2 in the schema URL

src/main/resources/org/springframework/integration/config/xml/spring-integration-kafka-1.2.xsd
```diff
@@ -117,6 +141,18 @@
 											</xsd:appinfo>
 										</xsd:annotation>
 									</xsd:attribute>
+									<xsd:attribute name="value-serializer" use="optional" type="xsd:string">
+										<xsd:annotation>
+											<xsd:appinfo>
+												<xsd:documentation>
+													Custom implementation of a Kafka Encoder for encoding message values.
```
mbogoevici: Wording - should refer to serializer, explain that options are mutually exclusive.

src/main/java/org/springframework/integration/kafka/config/xml/KafkaProducerContextParser.java
```diff
@@ -88,7 +111,7 @@ private void parseProducerConfigurations(Element topics, ParserContext parserCon
 			IntegrationNamespaceUtils.setValueIfAttributeDefined(producerMetadataBuilder, producerConfiguration,
 					"async");
```
mbogoevici: Remove, not supported anymore.

src/main/resources/org/springframework/integration/config/xml/spring-integration-kafka-1.2.xsd
```diff
@@ -105,6 +105,30 @@
...
+												<xsd:documentation>
+													Class type used for the value
+												</xsd:documentation>
+												<tool:annotation kind="direct">
+													<tool:expected-type type="java.lang.Class"/>
+												</tool:annotation>
+											</xsd:appinfo>
+										</xsd:annotation>
+									</xsd:attribute>
 									<xsd:attribute name="value-encoder" use="optional" type="xsd:string">
```
garyrussell: Given that we have some major schema incompatibilities anyway, why do we need to continue to support Encoder?

Seems like we can still provide the adapter as a helper for users who have custom encoders but it's not unreasonable to have them do the wrapping themselves.

src/test/java/org/springframework/integration/kafka/listener/AbstractMessageListenerContainerTests.java
```diff
@@ -29,6 +29,7 @@
 import java.util.ArrayList;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
```
garyrussell: Java 8 dep.

src/test/java/org/springframework/integration/kafka/offset/AbstractOffsetManagerTests.java
```diff
@@ -277,13 +286,11 @@ protected DefaultConnectionFactory createConnectionFactory() throws Exception {
...
 
 	protected Producer<String, String> createProducer() {
-		Properties properties = TestUtils.getProducerConfig(kafkaRule.getBrokersAsString(),
-				TestPartitioner.class.getCanonicalName());
-		properties.put("serializer.class", StringEncoder.class.getCanonicalName());
-		properties.put("key.serializer.class", StringEncoder.class.getCanonicalName());
-		ProducerConfig producerConfig = new ProducerConfig(properties);
-		return new Producer<String, String>(new kafka.producer.Producer<String, String>(producerConfig));
-	}
+		Properties properties = new Properties();
```
garyrussell: Similar fix needed in `KafkaMessageDrivenChannelAdapterExternalTests`

src/test/java/org/springframework/integration/kafka/offset/AbstractOffsetManagerTests.java
```diff
@@ -277,13 +286,11 @@ protected DefaultConnectionFactory createConnectionFactory() throws Exception {
...
+		properties.setProperty("bootstrap.servers", kafkaRule.getBrokersAsString());
+		EncoderAdaptingSerializer<String> serializer = new EncoderAdaptingSerializer<>(new StringEncoder());
+		return new KafkaProducer<String, String>(properties, serializer, serializer);
+		}
```
garyrussell: too many tabs :smile:

src/test/java/org/springframework/integration/kafka/offset/AbstractOffsetManagerTests.java
garyrussell: I removed that comment - it was a local file :smile:

garyrussell:
Is there anything we can do to get rid of these...

>13:37:04.668 WARN  [main][org.apache.kafka.clients.producer.ProducerConfig] The configuration value.serializer = class org.springframework.integration.kafka.util.EncoderAdaptingSerializer was supplied but isn't a known config.
>13:37:04.668 WARN  [main][org.apache.kafka.clients.producer.ProducerConfig] The configuration key.serializer = class org.springframework.integration.kafka.util.EncoderAdaptingSerializer was supplied but isn't a known config.

?

garyrussell:
I ran some integration tests; all seems to work ok.

src/main/resources/org/springframework/integration/config/xml/spring-integration-kafka-1.2.xsd
mbogoevici: Generally speaking, I've tried to preserve backwards compatibility with regards to the schema, unless it was clearly impossible to do so, or it would have made things confusing/dangerous (e.g. `async`, or switching batch size from number of messages to bytes), and it seemed like in this case there wouldn't be a lot of extra effort to do so, but we can just drop support for that, too.

mbogoevici:
wrt. the warnings, I'll check.

mbogoevici:
Actually, the warnings are because of a bug in Kafka. When a `Serializer` instance is set explicitly, Kafka will also populate the configuration map behind the scenes as well (to meet some validation requirements). However, since the key is not used (the directly configured instance is set on the producer), it's reported as an 'unknown' configuration.  I don't think that there's a safe way to disable it without disabling other, legitimate warnings.

mbogoevici:
Added some fixes, rebased.

src/main/java/org/springframework/integration/kafka/config/xml/KafkaProducerContextParser.java
```diff
@@ -73,22 +79,52 @@ private void parseProducerConfigurations(Element topics, ParserContext parserCon
...
-					"key-class-type");
-			IntegrationNamespaceUtils.setValueIfAttributeDefined(producerMetadataBuilder, producerConfiguration,
-					"value-class-type");
+			producerMetadataBuilder.addConstructorArgValue(producerConfiguration.getAttribute("key-class-type"));
+			producerMetadataBuilder.addConstructorArgValue(producerConfiguration.getAttribute("value-class-type"));
+
+			String keySerializer = producerConfiguration.getAttribute("key-serializer");
+			String keyEncoder = producerConfiguration.getAttribute("key-encoder");
+			Assert.isTrue((StringUtils.hasText(keySerializer) ^ StringUtils.hasText(keyEncoder)),
+					"Exactly one of 'key-serializer' or 'key-encoder' can be specified");
```
garyrussell: s/can/must/ ?

src/main/java/org/springframework/integration/kafka/listener/ConcurrentMessageListenerDispatcher.java
```diff
@@ -100,6 +104,8 @@ public void stop(int stopTimeout) {
 				this.running = false;
 				delegates.flip().keyBag().toSet().forEachWith(stopDelegateProcedure, stopTimeout);
 			}
+			this.taskExecutor = null;
```
garyrussell: Need a setter for the TE in case they're running in a JEE container; also we can't null it if set by the user.

src/main/java/org/springframework/integration/kafka/config/xml/KafkaProducerContextParser.java
```diff
@@ -65,7 +71,7 @@ protected boolean isEligibleAttribute(String attributeName) {
 	}
 
 	private void parseProducerConfigurations(Element topics, ParserContext parserContext,
-			BeanDefinitionBuilder builder, Element parentElem) {
+																					 BeanDefinitionBuilder builder, Element parentElem) {
```
garyrussell: fewer tabs

src/main/java/org/springframework/integration/kafka/support/ProducerConfiguration.java
```diff
@@ -45,63 +44,70 @@
...
 
-	private static boolean isRawByteArray(final Object obj) {
-		return obj instanceof byte[];
+	private K convertKeyIfNecessary(Object messageKey) {
+		if (messageKey != null) {
+			if (getProducerMetadata().getKeyClassType().isAssignableFrom(messageKey.getClass())) {
+				return getProducerMetadata().getKeyClassType().cast(messageKey);
+			}
+			return conversionService.convert(messageKey, producerMetadata.getKeyClassType());
+		} else {
```
garyrussell: newline

src/main/java/org/springframework/integration/kafka/util/EncoderAdaptingSerializer.java
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright ${today.year} the original author or authors.
```
garyrussell: 2015

src/main/resources/META-INF/spring.schemas
garyrussell: Still needs fixing

src/main/resources/org/springframework/integration/config/xml/spring-integration-kafka-1.2.xsd
garyrussell: ditto

src/test/java/org/springframework/integration/kafka/listener/AbstractBrokerTests.java
```diff
@@ -162,4 +166,30 @@ private static void sleep(int time) {
...
+		}
+
+		public void send(Collection<ProducerRecord<K,V>> records) {
+			Future<RecordMetadata> lastFuture = null;
+			for (ProducerRecord<K, V> record : records) {
+					lastFuture = producer.send(record);
+			}
+			if (lastFuture != null) {
+				try {
+					lastFuture.get();
```
garyrussell: Should we do any validation?

src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java
```diff
@@ -70,15 +77,19 @@ protected void onInit() throws Exception {
 
 	@Override
 	protected void handleMessageInternal(final Message<?> message) throws Exception {
-		String topic = this.topicExpression != null
-				? this.topicExpression.getValue(this.evaluationContext, message, String.class)
+		String topic = this.topicExpression != null ?
+				this.topicExpression.getValue(this.evaluationContext, message, String.class)
 				: message.getHeaders().get(KafkaHeaders.TOPIC, String.class);
 
+		Integer partitionId = this.partitionExpression != null ?
```
garyrussell: We typically do things like this by setting the default expression to `"headers[" + KafkaHeaders.PARTITION_ID + "']"`, avoiding the ternary expression.

garyrussell:
That's all; we do need an update to the README, though.

src/test/java/org/springframework/integration/kafka/listener/AbstractBrokerTests.java
mbogoevici: Not sure - It's a test class, really - the only goal here was to make the test asynchronous. The null check is for the case when the record list is empty, for whatever reason.

src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java
mbogoevici: I just followed the existing pattern :). However, I think the current form is, in a way, better, since it avoids a SpEL evaluation if an expression is not actually set. I think that getting the header from the map is generally more efficient.

src/main/java/org/springframework/integration/kafka/config/xml/KafkaProducerContextParser.java
mbogoevici: Yes.

src/test/java/org/springframework/integration/kafka/listener/AbstractBrokerTests.java
garyrussell: ok; np; perhaps a comment to that effect would help other readers.

mbogoevici:
Just realized that this will need rebasing. I can do it if it's the only outstanding change.

garyrussell:
rebased, squashed, merged as 0a34e540c6788a1d7d0609e7f73664f354cc56ae

PR #36, opened 2015-05-22
Add Reactor based ringbuffer

mbogoevici:
restart support

blocking stop

Test fixes

Ensure that an executor can be set externally

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,6 +8,7 @@ apply plugin: 'idea'
 group = 'org.springframework.integration'
 
 repositories {
+        mavenLocal()
 	maven {
 		url 'https://repository.apache.org/content/groups/public'
 	}
@@ -23,6 +24,7 @@ ext {
 	kafkaVersion = '0.8.2.1'
 	metricsVersion = '2.2.0'
 	scalaVersion = '2.10'
+	reactor2Version = '2.0.2.RELEASE'
 	springIntegrationVersion = '4.1.3.RELEASE'
 
 	idPrefix = 'kafka'
@@ -57,6 +59,7 @@ dependencies {
 	compile "org.apache.avro:avro:$avroVersion"
 	compile "org.apache.avro:avro-compiler:$avroVersion"
 	compile "com.goldmansachs:gs-collections:$gsCollectionsVersion"
+	compile "io.projectreactor:reactor-core:$reactor2Version"
 
 	runtime "com.yammer.metrics:metrics-core:$metricsVersion"
 	runtime "com.yammer.metrics:metrics-annotation:$metricsVersion"
--- a/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
+++ b/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
@@ -16,8 +16,10 @@
 
 package org.springframework.integration.kafka.inbound;
 
-import java.util.HashMap;
 import java.util.Map;
+
+import kafka.serializer.Decoder;
+import kafka.serializer.DefaultDecoder;
 
 import org.springframework.integration.context.OrderlyShutdownCapable;
 import org.springframework.integration.endpoint.MessageProducerSupport;
@@ -31,11 +33,11 @@ import org.springframework.integration.support.DefaultMessageBuilderFactory;
 import org.springframework.integration.support.MessageBuilderFactory;
 import org.springframework.integration.support.MutableMessageBuilderFactory;
 import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.support.MessageBuilder;
+import org.springframework.messaging.support.MessageHeaderAccessor;
 import org.springframework.util.Assert;
-
-import kafka.serializer.Decoder;
-import kafka.serializer.DefaultDecoder;
 
 /**
  * @author Marius Bogoevici
@@ -55,6 +57,7 @@ public class KafkaMessageDrivenChannelAdapter extends MessageProducerSupport imp
 	private boolean useMessageBuilderFactory = false;
 
 	private boolean autoCommitOffset = true;
+	private MessageChannel outputChannel;
 
 	public KafkaMessageDrivenChannelAdapter(KafkaMessageListenerContainer messageListenerContainer) {
 		Assert.notNull(messageListenerContainer);
@@ -121,6 +124,12 @@ public class KafkaMessageDrivenChannelAdapter extends MessageProducerSupport imp
 	}
 
 	@Override
+	public void setOutputChannel(MessageChannel outputChannel) {
+		super.setOutputChannel(outputChannel);
+		this.outputChannel = outputChannel;
+	}
+
+	@Override
 	protected void onInit() {
 		this.messageListenerContainer.setMessageListener(autoCommitOffset ?
 				new AutoAcknowledgingChannelForwardingMessageListener()
@@ -168,7 +177,7 @@ public class KafkaMessageDrivenChannelAdapter extends MessageProducerSupport imp
 
 		@Override
 		public void doOnMessage(Object key, Object payload, KafkaMessageMetadata metadata) {
-			KafkaMessageDrivenChannelAdapter.this.sendMessage(toMessage(key, payload, metadata, null));
+			KafkaMessageDrivenChannelAdapter.this.outputChannel.send(toMessage(key, payload, metadata, null));
 		}
 
 	}
@@ -184,6 +193,7 @@ public class KafkaMessageDrivenChannelAdapter extends MessageProducerSupport imp
 		@Override
 		public void doOnMessage(Object key, Object payload, KafkaMessageMetadata metadata,
 				Acknowledgment acknowledgment) {
+
 			KafkaMessageDrivenChannelAdapter.this.sendMessage(toMessage(key, payload, metadata, acknowledgment));
 		}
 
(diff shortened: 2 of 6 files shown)


build.gradle
```diff
@@ -23,6 +24,7 @@ ext {
 	kafkaVersion = '0.8.2.1'
 	metricsVersion = '2.2.0'
 	scalaVersion = '2.10'
+	reactor2Version = '2.0.2.RELEASE'
```
garyrussell: Bump to 2.0.3 for merge.

src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
```diff
@@ -168,7 +177,7 @@ public AutoAcknowledgingChannelForwardingMessageListener() {
 
 		@Override
 		public void doOnMessage(Object key, Object payload, KafkaMessageMetadata metadata) {
-			KafkaMessageDrivenChannelAdapter.this.sendMessage(toMessage(key, payload, metadata, null));
+			KafkaMessageDrivenChannelAdapter.this.outputChannel.send(toMessage(key, payload, metadata, null));
```
artembilan: With this we don't get so much gain and lose here the `error-channel` flow if that. Plus `trackMessage` within the `MessageHistory` is standard SI feature.

src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
```diff
@@ -192,36 +202,29 @@ public void doOnMessage(Object key, Object payload, KafkaMessageMetadata metadat
...
 
 		if (this.useMessageBuilderFactory) {
 			return getMessageBuilderFactory()
 					.withPayload(payload)
-					.copyHeaders(headers)
+					.copyHeaders(headerAccessor.toMessageHeaders())
 					.build();
 		}
 		else {
-			return new KafkaMessage(payload, headers);
```
artembilan: This is redundant already. So, I'm going to remove them all on merge.

src/main/java/org/springframework/integration/kafka/listener/QueueingMessageListenerInvoker.java
```diff
@@ -78,14 +95,8 @@ public void enqueue(KafkaMessage message) {
 			// handle the case when the thread is interrupted while the adapter is still running
 			// retry adding the message to the queue until either we succeed, or the adapter is stopped
 			while (!added && this.running) {
-				try {
-					this.messages.put(message);
-					added = true;
-				}
-				catch (InterruptedException e) {
-					// we ignore the interruption signal if we are still running, but pass it on if we are stopped
-					wasInterruptedWhileRunning = true;
```
artembilan: Since we don't catch `InterruptedException` any more, `wasInterruptedWhileRunning` variable is redundant. Removing...

src/main/java/org/springframework/integration/kafka/listener/QueueingMessageListenerInvoker.java
```diff
@@ -78,14 +95,8 @@ public void enqueue(KafkaMessage message) {
...
-				}
+				ringBufferProcessor.onNext(message);
+				added = true;
```
artembilan: The `added` here is redundant as well. Any exception in the `ringBufferProcessor` is propagated to the `Subscriber.onError`

src/main/java/org/springframework/integration/kafka/listener/QueueingMessageListenerInvoker.java
```diff
@@ -94,63 +105,69 @@ public void enqueue(KafkaMessage message) {
 	}
 
 	public void start() {
+
+		if(ringBufferProcessor != null){
+			stop(-1l);
```
artembilan: Frankly: I don't like this code. `start()` must be *idempotent*. And only the upstream `stop()` should reset the state. So, let me revert this dangerous code!

artembilan:
Merged as https://github.com/spring-projects/spring-integration-kafka/commit/74bb9b3cc05019ae28d986b115ccdea059dfe68d

src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
markpollack: I'm wondering now if this made a perf impact.... It may have...

markpollack: As we kept having to avoid using message builder often to increase perf

markpollack: As message builder spends time copying values which is avoided when using the KafkaMessage subclass.

artembilan: @markpollack , in this case no, it doesn't copy values. This one is `org.springframework.messaging.support.MessageBuilder`, but not from SI.

I think, when @mbogoevici was doing this change, he checked the performance.

markpollack: We aren't seeing the expected perf, will investigate today with a local build and this change.

markpollack: It makes a copy, in https://github.com/spring-projects/spring-framework/blob/master/spring-messaging/src/main/java/org/springframework/messaging/MessageHeaders.java. It creates a new map from an existing one.. This is called via new GenericMessage.  I remember us going down this path when refactoring last Thursday but will verify.

artembilan: M-m-m. @markpollack , the pattern looks like:
```
MessageHeaderAccessor headerAccessor = new MessageHeaderAccessor();
```
Which does:
```
public MessageHeaderAccessor() {
	this.headers = new MutableMessageHeaders();
}
```
Than we add desired headers like this:
```
headerAccessor.setHeader(KafkaHeaders.TOPIC, metadata.getPartition().getTopic());
```
And before message creation:
```
headerAccessor.setLeaveMutable(!(this.generateMessageId || this.generateTimestamp));
```
Which impacts to this:
```
return MessageBuilder.createMessage(payload, headerAccessor.getMessageHeaders());
```
where we have:
```
return new GenericMessage<T>(payload, messageHeaders);
```
just with this:
```
public GenericMessage(T payload, MessageHeaders headers) {
	Assert.notNull(payload, "Payload must not be null");
	Assert.notNull(headers, "MessageHeaders must not be null");
	this.payload = payload;
	this.headers = headers;
}
```

Have I missed anything?

If you do a performance testing from XD, maybe it ends up with this branch of logic:
```
if (this.useMessageBuilderFactory) {
	return getMessageBuilderFactory()
			.withPayload(payload)
			.copyHeaders(headerAccessor.toMessageHeaders())
			.build();
}
```
Where we really do the copy.
Need to know Kafka source config on the matter.

From other side we really might broke some logic, when the best part had been by default before.

markpollack: Yes, I think we are not going down the right branch, will modify the KafkaMessageBus in XD and see what happens.  I was trying to follow the calling chain on my iphone... Thought we were passing in a Map to create the GenericMessage, which results in creating a new HashMap.  Why do we do

````
	protected MessageHeaders(Map<String, Object> headers, UUID id, Long timestamp) {
		this.headers = (headers != null ? new HashMap<String, Object>(headers) : new HashMap<String, Object>());
````
vs just doing the assignment of the passed in headers?  In these type of micro benchmarks the extra copying does come into play.  Separate from this Kafka discussion..

File: src/main/java/org/springframework/integration/kafka/listener/QueueingMessageListenerInvoker.java (hunk 5 above)
Comment (mbogoevici) [inline src/main/java/org/springframework/integration/kafka/listener/QueueingMessageListenerInvoker.java]: Actually, this is wrong. The thread calling onNext(message) will block if the buffer is full, therefore the previous way of handling interruptions must be reinstated. Will issue a PR.

File: src/main/java/org/springframework/integration/kafka/listener/QueueingMessageListenerInvoker.java [hunk 7]
```diff
@@ -94,63 +105,69 @@ public void enqueue(KafkaMessage message) {
...
 	public void stop(long stopTimeout) {
-		shutdownLatch = new CountDownLatch(1);
 		this.running = false;
-		try {
-			shutdownLatch.await(stopTimeout, TimeUnit.MILLISECONDS);
-		}
-		catch (InterruptedException e) {
-			Thread.currentThread().interrupt();
+		if(ringBufferProcessor != null) {
+			ringBufferProcessor.onComplete();
```
mbogoevici: Shutdown latch must be set before `onComplete` or else we risk blocking needlessly below.

src/main/java/org/springframework/integration/kafka/listener/QueueingMessageListenerInvoker.java
artembilan: So, what you are saying must looks like:
````java
public void enqueue(KafkaMessage message) {
	boolean wasInterruptedWhileRunning = false;
	if (this.running) {
		boolean added = false;
		// handle the case when the thread is interrupted while the adapter is still running
		// retry adding the message to the queue until either we succeed, or the adapter is stopped
		while (!added && this.running) {
			try {
				ringBufferProcessor.onNext(message);
				added = true;
			}
			catch (InterruptedException e) {
				// we ignore the interruption signal if we are still running, but pass it on if we are stopped
				wasInterruptedWhileRunning = true;
			}
		}
	}
	if (wasInterruptedWhileRunning) {
		Thread.currentThread().interrupt();
	}
}
````
Don't you want to say that this is a typical pattern to get deal with `RingBuffer` ?
The blocking code looks like:
````java
while (wrapPoint > (minSequence = Util.getMinimumSequence(gatingSequences, nextValue))) {
      LockSupport.parkNanos(1L); // TODO: Use waitStrategy to spin
}
````
So, it isn't *blocking* really...

Have I missed anything?

src/main/java/org/springframework/integration/kafka/listener/QueueingMessageListenerInvoker.java
artembilan: Yeah... My bad on review.
Feel free to raise a fix PR!

src/main/java/org/springframework/integration/kafka/listener/QueueingMessageListenerInvoker.java
mbogoevici: Actually, you're right. `onNext` is non-blocking.

PR #41, opened 2015-06-01
QueueingMessageListenerInvoker fixes

mbogoevici:
- create latch before initiating shutdown;
- do not set latch to null inside onComplete();
- synchronize start and stop;


artembilan:
LGTM! :+1:

PR #42, opened 2015-06-02
INTEXT-170 Use a custom MessageHeaders class

mbogoevici:
- revert to using a custom KafkaMessageHeaders class directly;
- avoid the use of MutableMessageHeaders and the subsequent assignment of ids and timestamps within a GenericMessageTemplate


garyrussell:
Rebased, fixed unused imports; merged.

PR #43, opened 2015-06-08
INTEXT-172 Make potentially long-running tasks SchedulingAwareRunnable

mbogoevici:


mbogoevici:
UpdateOffsetsTask is not set set to be `SchedulingAwareRunnable` as it's a simple, short lived async task.

PR #46, opened 2015-06-18
INTEXT-175 Replace onInit() check for IntegrationEvaluationContext

mbogoevici:
- replace the `onInit()` check for an `IntegrationEvaluationContext` with a getter


artembilan:
Nothing to complain about!
Feel free to merge it yourself since SIK is open for you.
Otherwise I'd prefer to rework SI according to the mentioned JIRA and don't do nothing here and in other similar places, like Splunk, Cassandra, AWS etc...

mbogoevici:
Cool, thanks. I'll merge for now and revisit later - it's a simple way to ensure that the Spring IO plugin will work.

PR #52, opened 2015-07-17
INTEXT-180 Fetch pool size should be based on the number of brokers

mbogoevici:


artembilan:
Merging as is after some internal discussion

artembilan:
Merged as https://github.com/spring-projects/spring-integration-kafka/commit/d91e2efe704015d0e04e8981808ad7379dd6b31f after rebase

PR #58, opened 2015-08-27
Kafka native offset implementation

clemper:
This implementation uses kafka.client.ClientUtils.channelToOffsetManager to get the current kafka OffsetManager for the specified Partition.  It would probably be better to implement this logic directly in Java in DefaultConnectionFactory (i.e. getOffsetManager(parition) or similar) where the connection can be cached in a more thread-safe manner.

### DIFF
--- a/src/main/java/org/springframework/integration/kafka/listener/KafkaNativeOffsetManager.java
+++ b/src/main/java/org/springframework/integration/kafka/listener/KafkaNativeOffsetManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 the original author or authors.
+ * Copyright 2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -52,7 +52,7 @@ import kafka.utils.ZKStringSerializer$;
 /**
  * Implementation of an {@link OffsetManager} that uses kafka native 'topic' offset storage.
  *
- * @author clemper
+ * @author Chris Lemper
  */
 public class KafkaNativeOffsetManager extends AbstractOffsetManager implements InitializingBean {
 
--- a/src/main/java/org/springframework/integration/kafka/listener/KafkaNativeOffsetManager.java
+++ b/src/main/java/org/springframework/integration/kafka/listener/KafkaNativeOffsetManager.java
@@ -1,0 +1,235 @@
+/*
+ * Copyright 2014 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


artembilan:
Hello @clemper !

Thank you for the contribution, although we must be sure that you have singled the CLA already: https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.md.

I see that linked XD issue and really hope that this your fix will help not only to you :smile: .

From the subject of the PR I understand what you have done, but your comment isn't clear for me.
Would you mind to elaborate a bit more and maybe it really would be better to represent your idea as a code in the PR?

Thank you anyway!

clemper:
@artembilan 

I tried signing the CLA but it says I already signed it.  I did sign it for Spring XD previously.  

To elaborate on my comment, this implementation of OffsetManager uses the native kafka OffsetManager implementation instead of the custom offset topic implementation in KafkaTopicOffsetManager.  It uses as much of the existing support that was already in DefaultConnection as possible; however, you must execute the offset requests against the "offset manager coordinator" broker for the partition and there was no support for this already in spring-integration-kafka.  It turns out that the logic for finding the finding the offset manager broker is fairly complicated as described [here](https://cwiki.apache.org/confluence/display/KAFKA/Committing+and+fetching+consumer+offsets+in+Kafka).  Instead of implementing this in Java directly I call ClientUtils.channelToOffsetManager in the function KafkaNativeOffsetManager.getOffsetManagerBrokerAddress().  This is really a hack and requires creating a kafka BlockingChannel to the broker which I immediately close after getting the host and port of the broker.  I think a better solution would be to add this logic directly to DefaultConnectionFactory since it already has connection caching and functions like getLeader(partition).  This would simplify KafkaNativeOffsetManager and would allow getting the offset manager and executing the offset requests directly.  

Note that to get the offset manager in Java you would have to use the low level BlockingChannel since the SimpleConsumer java api does not expose the ConsumerMetadataRequest function.

mbogoevici:
Hi Chris,

Thanks for the contribution.

I don't see any particular issue with enhancing ConnectionFactory as a way to simplify this implementation. We should probably try that. 

Cheers,
Marius

src/main/java/org/springframework/integration/kafka/listener/KafkaNativeOffsetManager.java
```diff
@@ -0,0 +1,235 @@
...
+import kafka.client.ClientUtils$;
+import kafka.common.ErrorMapping;
+import kafka.common.OffsetAndMetadata;
+import kafka.network.BlockingChannel;
+import kafka.utils.ZKStringSerializer$;
+
+/**
+ * Implementation of an {@link OffsetManager} that uses kafka native 'topic' offset storage.
+ *
+ * @author clemper
```
mbogoevici: Chris, can you provide your full name here, please?

src/main/java/org/springframework/integration/kafka/listener/KafkaNativeOffsetManager.java
```diff
@@ -0,0 +1,235 @@
+/*
+ * Copyright 2014 the original author or authors.
```
mbogoevici: Needs to be 2015

mbogoevici:
Looks good, will factor out offset manager retrieval to ConnectionFactory

mbogoevici:
Closing, opened https://github.com/spring-projects/spring-integration-kafka/pull/79 to add the extra refactoring.

PR #68, opened 2015-09-24
convertPayloadIfNecessary change to use Value Class Type

gmateo:
The `convertPayloadIfNecessary` is using `getProducerMetadata().getKeyClassType()` when it should use `getProducerMetadata().getValueClassType()`

### DIFF
--- a/src/main/java/org/springframework/integration/kafka/support/ProducerConfiguration.java
+++ b/src/main/java/org/springframework/integration/kafka/support/ProducerConfiguration.java
@@ -103,7 +103,7 @@ public class ProducerConfiguration<K, V> {
 
 	private V convertPayloadIfNecessary(Object messagePayload) {
 		if (messagePayload != null) {
-			if (getProducerMetadata().getKeyClassType().isAssignableFrom(
+			if (getProducerMetadata().getValueClassType().isAssignableFrom(
 					messagePayload.getClass())) {
 				return getProducerMetadata().getValueClassType().cast(messagePayload);
 			}


artembilan:
Thank you for the contribution!
Would you mind following with the https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.md and confirm that you have signed the CLA before we start to take a look to your changes?

gmateo:
I have signed and agree to the terms of the SpringSource Individual
Contributor License Agreement.

artembilan:
Good.

Would you mind proving the issue with some short test-case?
If we'd have it before, we wouldn't faced out this ticket at all...

Thank you again!

gmateo:
Artem,

 

               We have had problems using the Kafka producer and then tracing to find the error we found this method:

 

private V convertPayloadIfNecessary(Object messagePayload) {
   if (messagePayload != null) {
      if (getProducerMetadata().getKeyClassType().isAssignableFrom(
            messagePayload.getClass())) {
         return getProducerMetadata().getValueClassType().cast(messagePayload);
      }
      return this.conversionService.convert(messagePayload, this.producerMetadata.getValueClassType());
   }
   else {
      return null;
   }
}

 

               And in this you can see clearly that the key class type is used  to check the payload class and if it is true you are casting the payload to the value class type; it is not logical.

               Please check that code and comment.

 

               Regards

Gamal

 

De: Artem Bilan [mailto:notifications@github.com] 
Enviado el: jueves, 24 de septiembre de 2015 04:53 p.m.
Para: spring-projects/spring-integration-kafka
CC: Gamal
Asunto: Re: [spring-integration-kafka] convertPayloadIfNecessary change to use Value Class Type (#68)

 

Good.

Would you mind proving the issue with some short test-case?
If we'd have it before, we wouldn't faced out this ticket at all...

Thank you again!

artembilan:
Yes, I understand what's going on. I just want to have the proper contribution from you.
And one of the requirements is test-case to demonstrate the issue, even if it is so simple like this.

That's why I said it is bad that we haven't had the test before...

Anyway it would be better to start from the test-case to reproduce on my side to confirm that it is an issue. 

I showed you the link for Contribution Guidline. It is enough simple bunch of rules and steps to follow and provide a good contribution.

mbogoevici:
Merged w/https://github.com/spring-projects/spring-integration-kafka/commit/2db9d1890dd9b72a23953864c9c88e4d43d2d796.

Added tests via https://github.com/spring-projects/spring-integration-kafka/commit/3c840dbd480c41360b414c73622f2a39b05ce03a

mbogoevici:
Thanks for contributing @gmateo - I have merged it as is and added a test. We strongly encourage submitting supporting tests though (even for the most trivial of issues).

PR #72, opened 2015-10-16
Support for sending messages synchronously to the Kafka broker.

martindam:
Adds two fields to the producer-configuration that allows to set the send mode (async or sync) and an optional timeout in sync mode. Default behavior is async.
This allows for an upstream application to verify that the message have been handed off to Kafka following the delivery options specified, and not just residing in memory at the producer.

### DIFF
--- a/src/main/java/org/springframework/integration/kafka/support/ProducerMetadata.java
+++ b/src/main/java/org/springframework/integration/kafka/support/ProducerMetadata.java
@@ -49,6 +49,10 @@ public class ProducerMetadata<K,V> {
 
 	private int batchBytes = 16384;
 
+	private int sendTimeout = 0;
+
+	private boolean sync;
+
 	public ProducerMetadata(final String topic, Class<K> keyClassType, Class<V> valueClassType,
 	                        Serializer<K> keySerializer, Serializer<V> valueSerializer) {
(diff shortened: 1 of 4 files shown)


mbogoevici:
Hi Martin,

Thanks for the contribution. Please sign the CLA as explained here (if you didn't already).

Thanks,
Marius

src/main/java/org/springframework/integration/kafka/support/ProducerMetadata.java
```diff
@@ -49,6 +49,10 @@
 
 	private int batchBytes = 16384;
 
+	private int sendTimeout = 0;
+
+	private SendMode sendMode = SendMode.async;
```
mbogoevici: I think that this should be a boolean value (`sync`), since there are only two possible modes - also simplifies evaluation.

martindam:
Thanks for the feedback Marius. 
I changed as you suggested and signed the CLA, and as asked in the readme:

I have signed and agree to the terms of the SpringSource Individual Contributor License Agreement.

mbogoevici:
Merged w/minor updates here: https://github.com/spring-projects/spring-integration-kafka/commit/14110e3741722e09ab64848aa1d8da9c1b500941

PR #79, opened 2015-10-23
Native offset

mbogoevici:


src/main/java/org/springframework/integration/kafka/listener/KafkaNativeOffsetManager.java
```diff
@@ -0,0 +1,235 @@
...
+
+	private static final Logger LOGGER = LoggerFactory.getLogger(KafkaNativeOffsetManager.class);
+
+	private static final String PARTITION_ATTRIBUTE = "partition";
+
+	private final Map<Partition, BrokerAddress> offsetManagerBrokerAddressCache = new ConcurrentHashMap<>();
+
+	private RetryTemplate retryTemplate;
+
+	private ZkClient zkClient;
```
artembilan: Should be `final` since it is initialized from the ctor

src/main/java/org/springframework/integration/kafka/listener/KafkaNativeOffsetManager.java
```diff
@@ -0,0 +1,235 @@
...
+	 * @param connectionFactory a Kafka connection factory
+	 * @param zookeeperConnect the zookeeper connection information
+	 * @param consumerId the kafka consumer ID
+	 * @param referenceTimestamp the reset timestamp for initial offsets
+	 */
+	public KafkaNativeOffsetManager(ConnectionFactory connectionFactory, ZookeeperConnect zookeeperConnect,
+									String consumerId, long referenceTimestamp) {
+		this(connectionFactory, zookeeperConnect, Collections.<Partition, Long>emptyMap());
+		setConsumerId(consumerId);
+		setReferenceTimestamp(referenceTimestamp);
```
artembilan: Minor, of course, but I wonder if we really should have these two options as ctor args.
Maybe that one for the `connectionFactory` and `zookeeperConnect` would be enough ?

src/main/java/org/springframework/integration/kafka/listener/KafkaNativeOffsetManager.java
```diff
@@ -0,0 +1,235 @@
...
+			backOffPolicy.setMaxInterval(5000L);
+			backOffPolicy.setMultiplier(2);
+			retryTemplate.setBackOffPolicy(backOffPolicy);
+		}
+	}
+
+	@Override
+	public void close() throws IOException {
+		// this function left intentionally blank
+		zkClient.close();
```
artembilan: I'm not sure now that is is still **blank**...

src/main/java/org/springframework/integration/kafka/listener/KafkaNativeOffsetManager.java
```diff
@@ -0,0 +1,235 @@
...
+	private class ResetOffsetManagerBrokerAddressRetryListener extends RetryListenerSupport {
+
+		@Override
+		public <T, E extends Throwable> void onError(RetryContext context, RetryCallback<T, E> callback, Throwable t) {
+			if (LOGGER.isWarnEnabled()) {
+				LOGGER.warn("Retrying kafka operation [" + callback + "] due to [" + t + "]", t);
+			}
+			final Object partition = context.getAttribute(PARTITION_ATTRIBUTE);
+			if (partition != null) {
+				offsetManagerBrokerAddressCache.remove(partition);
```
artembilan: Minor, of course, but it would be better to make our IDEs happy from those yellow lines :smile: 

suspicious for map remove

src/main/java/org/springframework/integration/kafka/listener/KafkaNativeOffsetManager.java
```diff
@@ -0,0 +1,235 @@
...
+
+/**
+ * Implementation of an {@link OffsetManager} that uses kafka native 'topic' offset storage.
+ *
+ * @author Chris Lemper
+ * @author Marius Bogoevici
+ */
+public class KafkaNativeOffsetManager extends AbstractOffsetManager implements InitializingBean {
+
+	private static final Logger LOGGER = LoggerFactory.getLogger(KafkaNativeOffsetManager.class);
```
artembilan: This is redundant. The super has `log` already

src/test/java/org/springframework/integration/kafka/offset/KafkaNativeOffsetManagerTests.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.springframework.integration.kafka.core.DefaultConnectionFactory;
+import org.springframework.integration.kafka.core.Partition;
+import org.springframework.integration.kafka.core.ZookeeperConfiguration;
+import org.springframework.integration.kafka.listener.KafkaNativeOffsetManager;
+import org.springframework.integration.kafka.listener.OffsetManager;
+import org.springframework.integration.kafka.support.ZookeeperConnect;
+
+/**
+ * @author clemper
```
artembilan: s/clemper/Chris Lemper
:smile:

artembilan:
Everything other is fine with me.
Just let me know - and I'll polish those simple concerns and merge!

src/main/java/org/springframework/integration/kafka/listener/KafkaNativeOffsetManager.java
mbogoevici: I was going to allow it like this, but on a second thought, you're right. Forcing the user to provide these arguments feels wrong.

mbogoevici:
@artembilan I commented on the constructor. Agreed on all the accounts, otherwise. I'll try to push some fixes tomorrow morning, but feel free to make the changes on merge.

artembilan:
@mbogoevici , since you haven't done anything here since the last our discussion, so I go ahead with our conclusion and fixing those concerns on merge.

Let me know if that should be some other way.

PR #80, opened 2015-10-24
Adds support for ProducerListener

mbogoevici:
Introduces the concept of ProducerListener, that can be injected in a ProducerContext and receive notifications when a message has been acknowledged or rejected.

While the Kafka API currently supports providing a Callback, it requires the creation of a distinct instance for receiving contextual information about the message that has been sent or rejected. The newly added feature allows the use of the Callback mechanism with a configurable, injectable ProducerListener strategy.

### DIFF
--- a/src/main/java/org/springframework/integration/kafka/support/LoggingProducerListener.java
+++ b/src/main/java/org/springframework/integration/kafka/support/LoggingProducerListener.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.kafka.support;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.springframework.util.ObjectUtils;
+
+/**
+ * {@link ProducerListener} that logs exceptions thrown when sending messages.
+ *
+ * @author Marius Bogoevici
+ * @since 1.3
+ */
+public class LoggingProducerListener extends DefaultProducerListener {
+
+	private static final Log log = LogFactory.getLog(LoggingProducerListener.class);
+
+	private boolean includeContents = true;
+
+	private int maxContentLogged = 100;
+
+	/**
+	 * Whether the log message should include the contents (key and payload).
+	 *
+	 * @param includeContents true if the contents of the message should be logged
+	 */
+	public void setIncludeContents(boolean includeContents) {
+		this.includeContents = includeContents;
+	}
+
+	/**
+	 * The maximum amount of data to be logged for either key or password. As message sizes may vary and
+	 * become fairly large, this allows limiting the amount of data sent to logs.
+	 *
+	 * @param maxContentLogged the maximum amount of data being logged.
+	 */
+	public void setMaxContentLogged(int maxContentLogged) {
+		this.maxContentLogged = maxContentLogged;
+	}
+
+	@Override
+	public void onError(String topic, Integer partition, Object key, Object payload, Exception exception) {
+		// key and payload are intentionally left out, due to security and/or size concerns
+		if (log.isErrorEnabled()) {
+			StringBuffer logOutput = new StringBuffer();
+			logOutput.append("Exception thrown when sending a message");
+			if (includeContents) {
+				logOutput.append(" with key='"
+						+ toDisplayString(ObjectUtils.nullSafeToString(key), maxContentLogged) + "'");
+				logOutput.append(" and payload='"
+						+ toDisplayString(ObjectUtils.nullSafeToString(payload),maxContentLogged) + "'");
+			}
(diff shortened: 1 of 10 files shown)


src/main/java/org/springframework/integration/kafka/support/LoggingProducerListener.java
```diff
@@ -0,0 +1,84 @@
...
+	 *
+	 * @param maxContentLogged the maximum amount of data being logged.
+	 */
+	public void setMaxContentLogged(int maxContentLogged) {
+		this.maxContentLogged = maxContentLogged;
+	}
+
+	@Override
+	public void onError(String topic, Integer partition, Object key, Object payload, Exception exception) {
+		// key and payload are intentionally left out, due to security and/or size concerns
```
garyrussell: Looks like this comment was left over from before `includeContents` was added?

mbogoevici: Correct. Will fix.

garyrussell:
Looks ok to me - it's a bit confusing though (the kafka API).

The `send`already returns a `Future` so, presumably you can get the `RecordMetadata` (and or exception) that way.

So, is this simply a mechanism to save the user from having to retain the context for the send and to get an async callback rather than calling `get()` on the `Future` ?

Would he simply discard the `Future` when using this mechanism?

I wonder if we should change the API to return a `ListenableFuture` and use it's callback mechanism.

i.e. `public ListenableFuture<RecordMetadata> send(...)`

It would be backward compatible for users who assign the result to a `Future`.

Then, internally, when the kafka callback is invoked we simply populate the `ListenableFuture` with the metadata; perhaps a subclass of `SettableListenableFuture`.

So the user's interaction (and result/error detection) is entirely centered around the `Future`.

mbogoevici:
Yes, this is essentially a mechanism for receiving an asynchronous callback. What it does is that it guarantees that Callbacks are invoked in the exact order in which messages are acknowledged (or rejected) by brokers, for a given target partition. 

The main problem that I was trying to solve was that the users of the `KafkaMessageListenerContainer` and then the `KafkaMessageDrivenChannelAdapter` do not have access to the `Future` itself. I prefer relying on Kafka's own `Callback` mechanism because it is natively provided, and second, because of the strong guarantee from the API that the callbacks will be invoked in the order in which messages are sent to a given partition - which is something we couldn't guarantee for a ListenableFuture - or if we could, it would require extra effort.

garyrussell:
ok; sounds reasonable; it just seems a bit odd that the send with the callback also returns a future, but it's not the first odd Kafka API we've seen :smile: Of course, their javadoc makes no mention of it.

I'll remove that comment and merge.

garyrussell:
Maybe we should add an integration test?

garyrussell:
@mbogoevici I added an [integration test](https://github.com/garyrussell/spring-integration-kafka/commit/45e4bf413b5c38528f8ab7d692d406aa1ff7a9cb) but I'm not seeing kafka invoking the callback.

Any ideas?

garyrussell:
FYI; I am seeing the callbacks ok with this: https://www.codatlas.com/github.com/apache/kafka/HEAD/examples/src/main/java/kafka/examples/Producer.java?line=1

Investigating the difference...

I also see it's not necessary for us to save off the `topic` and `partition` - they are returned in the `RecordMetadata`.

mbogoevici:
Yes, that is for the exception. No `RecordMetadata` is returned in that case.

src/main/java/org/springframework/integration/kafka/support/ProducerListenerInvokingCallback.java
```diff
@@ -0,0 +1,61 @@
...
+		this.topic = topic;
+		this.partition = partition;
+		this.key = key;
+		this.payload = payload;
+		this.producerListener = producerListener;
+	}
+
+	@Override
+	public void onCompletion(RecordMetadata metadata, Exception exception) {
+		if (exception == null) {
```
mbogoevici: This needs to be the other way around. `!=`

garyrussell: Doh - missed that - however, I set a breakpoint on that line in the debugger early on in my testing and it never got a hit.

mbogoevici: It does, there seems to be a longer timeout involved. I'll get back a little later with an update PR.

garyrussell:
Polishing and test here:https://github.com/garyrussell/spring-integration-kafka/commit/95b6c3e51a53785d9e1b811d02164194bd1aadd1 let me know if it's ok and I'll merge.

mbogoevici:
Hi @garyrussell - the extra commit looks good - thanks.

garyrussell:
Merged as 325d0b294c825252f5313bd9696229d198b0d38e

sharpnk:
Excuse me if this is not the best place to ask a question regarding ProducerListener. I am trying to figure out what's the best way to incorporate this ProducerListener in my project. In my use case, I need to send out a bunch of messages to Kafka broker in strict order. If one of those messages failed to be delivered to the broker, the caller needs to be notified and stop all subsequent requests. ProducerListener looks like a great candidate to solve my problem and guarantee the delivery order. 

I created a ProducerListener and assigned it to the producer configuration. The caller sends these messages to a MessageChannel which is wired up to a outbound-channel-adapter. Ideally, the caller should be notified of whether the message delivery to the broker is successful or not. However, because the ProducerListener and its callbacks are pre-defined with the producer configuration, I found myself stuck in a place where I can't get the call back to properly notify the caller. 

To go around this, I added a ConcurrentHashSet in the the ProducerListner and keep track of all the failed delivery attempt. This way the caller will be able to check it next time it tries to send messages. This is far from ideal though. 

Any suggestions on this? I am just curious if I am using it as it's intended to be. Thanks.

mbogoevici:
hi @sharpnk,

Please open an issue.

PR #81, opened 2015-10-27
INTEXT-206: Don't create extra consumers

artembilan:
JIRA: https://jira.spring.io/browse/INTEXT-206

The previous logic in the `ConcurrentMessageListenerDispatcher` create exactly `consumers`
number of `QueueingMessageListenerInvoker`, although we use only the `partitions.size()`.

Fix the logic to the `this.consumers = Math.min(partitions.size(), consumers);` to create maximum
consumers according to the partitions count.

**Cherry-pick to 1.2.x**


mbogoevici:
Merged w/https://github.com/spring-projects/spring-integration-kafka/commit/8b56a7dc3f4178d74b7581bb0826af93c5f66156

PR #82, opened 2015-10-29
INTEXT-139: Deprecate High Level Consumer Adapter

artembilan:
JIRA: https://jira.spring.io/browse/INTEXT-139


mbogoevici:
I don't think that we want to remove the tests for the deprecated classes. I think that it's still good to have them around until we remove those components altogether.

artembilan:
Yeah... That was an evening's overhead...
Changed the mind and pushed deprecation instead of removal.

mbogoevici:
Merged with minor changes: https://github.com/spring-projects/spring-integration-kafka/commit/0ab6eedc528b495da8a8db82de014bf80fe1ab60

PR #83, opened 2015-11-02
Add header routing setting to KafkaOutboundChannelAdapter

mbogoevici:


src/main/java/org/springframework/integration/kafka/config/xml/KafkaOutboundChannelAdapterParser.java
```diff
@@ -67,6 +67,9 @@ protected AbstractBeanDefinition parseConsumer(final Element element, final Pars
 			kafkaProducerMessageHandlerBuilder.addPropertyValue("partitionIdExpression", partitionIdExpressionDef);
 		}
 
+		IntegrationNamespaceUtils.setValueIfAttributeDefined(kafkaProducerMessageHandlerBuilder, element,
+				"enable-header-routing", "enableHeaderRouting");
```
artembilan: No reason to specify property, if it is can be fabricated from dash-style.
See `IntegrationNamespaceUtils` source code:
```
setValueIfAttributeDefined(builder, element, attributeName,
		Conventions.attributeNameToPropertyName(attributeName), emptyStringAllowed);
```

A `Conventions.attributeNameToPropertyName()` is the trick.

src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java
```diff
@@ -47,6 +50,20 @@ public KafkaProducerMessageHandler(final KafkaProducerContext kafkaProducerConte
 		this.kafkaProducerContext = kafkaProducerContext;
 	}
 
+	/**
+	 * Enables the use of headers for determining the target topic and partition of outbound messages. By default it is
```
artembilan: Method JavaDocs in the *imperative mood*

src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java
```diff
@@ -47,6 +50,20 @@ public KafkaProducerMessageHandler(final KafkaProducerContext kafkaProducerConte
...
+	 * set to true, but it can be disabled when those values are produced by upstream components that read messages
+	 * from Kafka sources themselves.
+	 *
+	 * @see KafkaHeaders#TOPIC
+	 * @see KafkaHeaders#PARTITION_ID
+	 * @param enableHeaderRouting whether the topic and destination headers should be considered
```
artembilan: Interesting point. We have other components with similar `expression -> headers` logic.
And for such an issue to avoid the incoming headers we always recommend to use `<header-filter>`.

I can guess that this approach would be better because we bypass extra `Message<?>` creation after the `<header-filter>`. 
From other side maybe we should just always ask for those `expressions` rather than fail with similar issues...
I mean don't fallback to the headers at all.

CC @garyrussell

src/test/java/org/springframework/integration/kafka/outbound/OutboundTests.java
```diff
@@ -276,6 +274,97 @@ public void testNoHeader() throws Exception {
...
+
+
+	@Test
+	public void testNoHeader() throws Exception {
+
+		// create the topic
+
+		try {
+			TopicUtils.ensureTopicCreated(kafkaRule.getZookeeperConnectionString(), TOPIC, 1, 1);
+		} catch (TopicExistsException e) {
```
artembilan: `catch` on new line.

mbogoevici:
Adressed PR comments.

src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java
mbogoevici: This would be possible too - I am just concerned that disabling headers altogether would be too much of a breaking change. I think that this strikes a pretty fair balance, also given that the `<header-filter>` solution may inadvertently remove headers that are not used in routing, e.g. `kafka_messageKey`.

artembilan: Yes, as a gold middle we can live with this idea as well. At least we can simply it include to the `1.3` and silently fix the XD issue just only with one more option.
We should add a note somewhere or `TODO` on those headers resolution.
And really remove this logic in the `2.0` in favor of the provided `expression`. Thus we are washing our hands and go to drink beer :smile:

mbogoevici: I agree with removing this for 2.0. All with the beer.

artembilan:
Sorry, I have missed this one somehow...
Grabbing for final review and merge.

src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java
artembilan: The JIRA https://jira.spring.io/browse/INTEXT-207 to consider.

artembilan:
Merged as https://github.com/spring-projects/spring-integration-kafka/commit/81e803365f67d609208551aab4aceb3de2a7c7aa

PR #87, opened 2015-11-09
Prevent the KafkaMessageListenerContainer from blocking during shutdown

mbogoevici:
If the RingBufferProcessor used by the KafkaMessageListenerContainer blocks during
processing, it may be prevented from handling the onConsume() notfication which is
the shutdown indicator. Thus, the stopping of the component may be effectively
delayed.

To solve this, the processing task of the RingBufferProcessor is wrapped in a
cancelable task that is managed by one of the container components, and canceled
on shutdown, allowing for the interruption of the processing thread. Subsequently,
all messages cached by the ringbuffer are discarded, without being processed or
acknowdledged.

Additionally, an option for acknowledging processed messages on success only is
added. This allows the component to run continuously and advance even in the case
of sporadic errors, but to resume from the last successfully processed message.

### DIFF
--- a/src/main/java/org/springframework/integration/kafka/listener/ConcurrentMessageListenerDispatcher.java
+++ b/src/main/java/org/springframework/integration/kafka/listener/ConcurrentMessageListenerDispatcher.java
@@ -68,9 +68,12 @@ class ConcurrentMessageListenerDispatcher {
 
 	private MutableMap<Partition, QueueingMessageListenerInvoker> delegates;
 
+	private boolean autoAckOnError;
+
 	public ConcurrentMessageListenerDispatcher(Object delegateListener, ErrorHandler errorHandler,
 											   Collection<Partition> partitions, OffsetManager offsetManager,
-											   int consumers, int queueSize, Executor taskExecutor) {
+											   int consumers, int queueSize, Executor taskExecutor,
+											   boolean autoAckOnError) {
 		Assert.isTrue
 				(delegateListener instanceof MessageListener
 								|| delegateListener instanceof AcknowledgingMessageListener,
@@ -87,6 +90,7 @@ class ConcurrentMessageListenerDispatcher {
 		this.consumers = Math.min(partitions.size(), consumers);
 		this.queueSize = queueSize;
 		this.taskExecutor = taskExecutor;
+		this.autoAckOnError = autoAckOnError;
 	}
 
 	public void start() {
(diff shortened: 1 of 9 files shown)


src/main/java/org/springframework/integration/kafka/listener/ConcurrentMessageListenerDispatcher.java
```diff
@@ -120,7 +124,7 @@ private void initializeAndStartDispatching() {
 		for (int i = 0; i < consumers; i++) {
 			QueueingMessageListenerInvoker queueingMessageListenerInvoker =
 					new QueueingMessageListenerInvoker(queueSize, offsetManager, delegateListener, errorHandler,
-							taskExecutor);
+							taskExecutor, autoAckOnError);
```
mbogoevici: Apparently the refactoring didn't pick this one up.

garyrussell: ? Do you mean commit Vs. ack?

mbogoevici: It's called `autoCommitOnError` in the `KafkaMessageListenerContainer` ... minor anyway.

garyrussell:
LGTM; I can fix those variable names on merge. Since `commit` is in the schema, I presume you want to use commit throughout.

Just so I understand - if a user wants to treat an error as fatal; his error-handler would have to stop the container and signal to his listener to reject all subsequent messages, to prevent the offset from being updated while draining the ring buffer - correct?

Then, when the container is restarted, messages will be redelivered from the first failure.

I wonder if we should consider a `stopOnFailure` boolean.

mbogoevici:
Thanks @garyrussell!

Just stopping the container is enough, as https://github.com/spring-projects/spring-integration-kafka/pull/87/files#diff-6e3a6d8ca8b454adcefa471fe099b46dR217 prevents processing new messages (and offset updates) after `stop()`.

Let's discuss that option of stopping on failure. I've considered it as part of the initial design, but it felt a little strange, seeing as we don't have any self-stopping message listeners (as far as I can tell).

It almost feels like the scenario when failures are so radical that they require stopping the entire processing flow are better suited for batch support, wdyt?

garyrussell:
I agree it's weird; however, I don't really see the point of `auto-commit-on-error` without stopping the container because the next good message will update the offset anyway.

mbogoevici:
Yes, that is more of a heuristic right now that's trying to distinguish between sporadic (a mix of successful and unsuccessful messages) and persistent failures (only unsuccessful messages). In the end, the component should be also be ready to be forcibly reset (via configuration) to a different offset than the one already set, so it could replay the listened topic(s) altogether.

garyrussell:
Squashed and merged as eada60c114fc842b38d0d435e22bc461d2538cbe with minor polish: https://github.com/garyrussell/spring-integration-kafka/commit/c249495545d12caaed817ecb3bf082e1a5dfe876

Issue #70, opened 2015-10-07
KafkaMessageDrivenChannelAdapter does not connect to new leaders


dlebrero:
We have noticed that the KafkaMessageDrivenChannelAdapter never refresh the list of brokers that it is connected to, which means that if:
* A Kafka broker is down when the Spring app starts, or
* A Kafka broker is added to the cluster after the Spring app starts, or
* A Kafka broker is not a leader of any partition when the Spring app starts

and the new Kafka broker become the leader of a partition, the Spring app will no consume messages from that partition.

Probably org.springframework.integration.kafka.listener.KafkaMessageListenerContainer.FetchTask.UpdateLeadersTask should create new FetchTasks for unknown/new brokers.

The easiest way to reproduce:
* Create a Kafka cluster with 2 brokers
* Create a topic with 1 partition
* Configure Spring to consume that topic
* Kill the leader broker.
* Spring stops consuming messages

mbogoevici:
Thanks @dlebrero. Failover should work at least now, but the initial list of brokers is not update. Will fix this in the next releases.

vladislav-sidorovich:
Probably fix https://github.com/spring-projects/spring-integration-kafka/pull/76

mbogoevici:
A quick update: we will support this as part of Spring Integration Kafka 1.3, to be released this week.

mbogoevici:
Hi,

I think https://github.com/spring-projects/spring-integration-kafka/commit/cabc9d14a717222fd6ffb3969288f44eba9b1d29 will fix this.

A few questions/comments about this:

- the instructions don't specify the replication factor when the topic has been created. I would assume that it's 2, but in case it is the default of 1 - then that is one reason why Spring Integration cannot consume from the topic - the Kafka cluster can't bring up a live leader;
- about adding a broker to the Kafka cluster - in order to become a leader, the broker must either have been alive when the partition has been created (so that a replica is assigned to it) - or must have been reassigned the partition while there is a live leader - Kafka will not replicate partitions automatically to newly arrived brokers.


Hope this helps,
Marius

dlebrero:
Hi Marius,

Thanks a lot. We will test it as soon as the new version is available.

Regards,

Daniel

adamdec:
Hi, is this issue already fixed in version 1.3.1.RELEASE or 2.0.1.RELEASE?

Thanks.

garyrussell:
[...]

PR #89, opened 2015-11-13
GH-70 Allow the client to discover new brokers

mbogoevici:
* Refactor the message listener container to finish tasks on broker connection failure
and spawn new ones whenever a new broker is discovered;
* Add an explicit disconnect operation on the ConnectionFactory to release the connection
on failure;
* Other improvements and test refactorings;

### DIFF
--- a/src/main/java/org/springframework/integration/kafka/core/DefaultConnectionFactory.java
+++ b/src/main/java/org/springframework/integration/kafka/core/DefaultConnectionFactory.java
@@ -169,9 +169,6 @@ public class DefaultConnectionFactory implements InitializingBean, ConnectionFac
 	public void refreshMetadata(Collection<String> topics) {
 		try {
 			this.lock.writeLock().lock();
-			for (Connection connection : this.kafkaBrokersCache) {
-				connection.close();
-			}
 			String brokerAddressesAsString =
 					ListIterate.collect(this.configuration.getBrokerAddresses(), Functions.getToString())
 							.makeString(",");
@@ -189,6 +186,20 @@ public class DefaultConnectionFactory implements InitializingBean, ConnectionFac
 					log.info(String.format("No metadata could be retrieved for '%s'", topicMetadata.topic()),
 							ErrorMapping.exceptionFor(topicMetadata.errorCode()));
 				}
+			}
+		}
+		finally {
+			this.lock.writeLock().unlock();
+		}
+	}
+
+	@Override
+	public void disconnect(BrokerAddress brokerAddress) {
+		try {
+			this.lock.writeLock().lock();
+			Connection connection = this.kafkaBrokersCache.get(brokerAddress);
+			if (connection != null) {
+				connection.close();
 			}
 		}
 		finally {
--- a/src/main/java/org/springframework/integration/kafka/listener/ConcurrentMessageListenerDispatcher.java
+++ b/src/main/java/org/springframework/integration/kafka/listener/ConcurrentMessageListenerDispatcher.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -63,6 +64,8 @@ class ConcurrentMessageListenerDispatcher {
 	private final int queueSize;
 
 	private final Executor taskExecutor;
+
+	private final AtomicInteger partitionAssignmentCounter = new AtomicInteger();
 
 	private volatile boolean running;
 
@@ -120,7 +123,7 @@ class ConcurrentMessageListenerDispatcher {
 
 	private void initializeAndStartDispatching() {
 		// allocate delegate instances index them
-		List<QueueingMessageListenerInvoker> delegateList = new ArrayList<QueueingMessageListenerInvoker>(consumers);
+		List<QueueingMessageListenerInvoker> delegateList = new ArrayList<>(consumers);
 		for (int i = 0; i < consumers; i++) {
 			QueueingMessageListenerInvoker queueingMessageListenerInvoker =
 					new QueueingMessageListenerInvoker(queueSize, offsetManager, delegateListener, errorHandler,
@@ -130,9 +133,8 @@ class ConcurrentMessageListenerDispatcher {
 
 		// evenly distribute partitions across delegates
 		delegates = Maps.mutable.of();
-		int i = 0;
 		for (Partition partition : partitions) {
-			delegates.put(partition, delegateList.get((i++) % consumers));
+			delegates.put(partition, delegateList.get((partitionAssignmentCounter.incrementAndGet()) % consumers));
 		}
 
 		// start dispatchers
--- a/src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
+++ b/src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
@@ -16,25 +16,29 @@
 
 package org.springframework.integration.kafka.listener;
 
-import static com.gs.collections.impl.utility.ArrayIterate.flatCollect;
-import static com.gs.collections.impl.utility.Iterate.partition;
-import static com.gs.collections.impl.utility.MapIterate.forEachKeyValue;
-
-import java.io.IOException;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.Executor;
-import java.util.concurrent.Executors;
-
+import com.gs.collections.api.RichIterable;
+import com.gs.collections.api.block.function.Function;
+import com.gs.collections.api.block.predicate.Predicate;
+import com.gs.collections.api.collection.MutableCollection;
+import com.gs.collections.api.list.ImmutableList;
+import com.gs.collections.api.list.MutableList;
+import com.gs.collections.api.multimap.list.ImmutableListMultimap;
+import com.gs.collections.api.multimap.set.MutableSetMultimap;
+import com.gs.collections.api.partition.PartitionIterable;
+import com.gs.collections.api.set.MutableSet;
+import com.gs.collections.api.tuple.Pair;
+import com.gs.collections.impl.block.factory.Functions;
+import com.gs.collections.impl.block.function.checked.CheckedFunction;
+import com.gs.collections.impl.factory.Lists;
+import com.gs.collections.impl.factory.Sets;
+import com.gs.collections.impl.list.mutable.FastList;
+import com.gs.collections.impl.map.mutable.UnifiedMap;
+import com.gs.collections.impl.utility.Iterate;
+import kafka.common.ErrorMapping;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-
 import org.springframework.context.SmartLifecycle;
+import org.springframework.core.task.SimpleAsyncTaskExecutor;
 import org.springframework.integration.kafka.core.BrokerAddress;
 import org.springframework.integration.kafka.core.ConnectionFactory;
 import org.springframework.integration.kafka.core.ConsumerException;
@@ -47,26 +51,18 @@ import org.springframework.integration.kafka.core.Partition;
 import org.springframework.integration.kafka.core.Result;
 import org.springframework.scheduling.SchedulingAwareRunnable;
 import org.springframework.util.Assert;
-import org.springframework.util.CollectionUtils;
 
-import com.gs.collections.api.RichIterable;
-import com.gs.collections.api.block.function.Function;
-import com.gs.collections.api.block.predicate.Predicate;
-import com.gs.collections.api.block.procedure.Procedure;
-import com.gs.collections.api.block.procedure.Procedure2;
-import com.gs.collections.api.collection.MutableCollection;
-import com.gs.collections.api.list.ImmutableList;
-import com.gs.collections.api.list.MutableList;
-import com.gs.collections.api.multimap.MutableMultimap;
-import com.gs.collections.api.partition.PartitionIterable;
-import com.gs.collections.impl.block.factory.Functions;
-import com.gs.collections.impl.block.function.checked.CheckedFunction;
-import com.gs.collections.impl.factory.Lists;
-import com.gs.collections.impl.factory.Multimaps;
-import com.gs.collections.impl.list.mutable.FastList;
-import com.gs.collections.impl.utility.Iterate;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
 
-import kafka.common.ErrorMapping;
+import static com.gs.collections.impl.utility.ArrayIterate.flatCollect;
+import static com.gs.collections.impl.utility.Iterate.partition;
 
 /**
  * @author Marius Bogoevici
@@ -87,8 +83,6 @@ public class KafkaMessageListenerContainer implements SmartLifecycle {
 
 	private final Function<Partition, Partition> passThru = Functions.getPassThru();
 
-	private final LaunchFetchTaskProcedure launchFetchTask = new LaunchFetchTaskProcedure();
-
 	private final Object lifecycleMonitor = new Object();
 
 	private final KafkaTemplate kafkaTemplate;
@@ -101,7 +95,7 @@ public class KafkaMessageListenerContainer implements SmartLifecycle {
 
 	private Executor fetchTaskExecutor;
 
-	private Executor adminTaskExecutor = Executors.newSingleThreadExecutor();
+	private Executor adminTaskExecutor;
 
 	private Executor dispatcherTaskExecutor;
 
@@ -125,7 +119,7 @@ public class KafkaMessageListenerContainer implements SmartLifecycle {
 
 	private ConcurrentMessageListenerDispatcher messageDispatcher;
 
-	private final MutableMultimap<BrokerAddress, Partition> partitionsByBrokerMap = Multimaps.mutable.set.with();
+	private final ConcurrentMap<BrokerAddress, FetchTask> fetchTasksByBroker = new ConcurrentHashMap<>();
 
 	private boolean autoCommitOnError = false;
 
@@ -159,11 +153,10 @@ public class KafkaMessageListenerContainer implements SmartLifecycle {
 	}
 
 	public void setMessageListener(Object messageListener) {
-		Assert.isTrue
-				(messageListener instanceof MessageListener
-								|| messageListener instanceof AcknowledgingMessageListener,
-						"Either a " + MessageListener.class.getName() + " or a "
-								+ AcknowledgingMessageListener.class.getName() + " must be provided");
+		Assert.isTrue(
+				messageListener instanceof MessageListener || messageListener instanceof AcknowledgingMessageListener,
+				"Either a " + MessageListener.class.getName() + " or a " + AcknowledgingMessageListener.class.getName()
+						+ " must be provided");
 		this.messageListener = messageListener;
 	}
 
@@ -180,8 +173,9 @@ public class KafkaMessageListenerContainer implements SmartLifecycle {
 	}
 
 	/**
-	 * The maximum number of concurrent {@link MessageListener}s running. Messages from within the same
-	 * partition will be processed sequentially.
+	 * The maximum number of concurrent {@link MessageListener}s running. Messages from
+	 * within the same partition will be processed sequentially.
+	 *
 	 * @param concurrency the concurrency maximum number
 	 */
 	public void setConcurrency(int concurrency) {
@@ -189,7 +183,9 @@ public class KafkaMessageListenerContainer implements SmartLifecycle {
 	}
 
 	/**
-	 * The timeout for waiting for each concurrent {@link MessageListener} to finish on stopping.
+	 * The timeout for waiting for each concurrent {@link MessageListener} to finish on
+	 * stopping.
+	 *
 	 * @param stopTimeout timeout in milliseconds
 	 * @since 1.1
 	 */
@@ -207,27 +203,29 @@ public class KafkaMessageListenerContainer implements SmartLifecycle {
 
 	/**
 	 * The task executor for fetch operations
+	 *
 	 * @param fetchTaskExecutor the Executor for fetch operations
 	 */
 	public void setFetchTaskExecutor(Executor fetchTaskExecutor) {
 		this.fetchTaskExecutor = fetchTaskExecutor;
 	}
 
-
 	public Executor getAdminTaskExecutor() {
 		return adminTaskExecutor;
 	}
 
 	/**
-	 * The task executor for leader and offset updates
-	 * @param adminTaskExecutor the task executor for leader and offset updates.
+	 * The task executor for leader, offset, and partition reassignment updates
+	 *
+	 * @param adminTaskScheduler the task executor for leader, offset and partition reassignment updates
 	 */
-	public void setAdminTaskExecutor(Executor adminTaskExecutor) {
-		this.adminTaskExecutor = adminTaskExecutor;
+	public void setAdminTaskExecutor(Executor adminTaskScheduler) {
+		this.adminTaskExecutor = adminTaskScheduler;
 	}
 
 	/**
 	 * The task executor for invoking the MessageListener
+	 *
 	 * @param dispatcherTaskExecutor the task executor for invoking the MessageListener
 	 */
 	public void setDispatcherTaskExecutor(Executor dispatcherTaskExecutor) {
@@ -246,9 +244,10 @@ public class KafkaMessageListenerContainer implements SmartLifecycle {
 	}
 
 	/**
-	 * The maximum number of messages that are buffered by each concurrent {@link MessageListener} runner.
-	 * Increasing the value may increase throughput, but also increases the memory consumption.
-	 * Must be a positive number and a power of 2.
+	 * The maximum number of messages that are buffered by each concurrent
+	 * {@link MessageListener} runner. Increasing the value may increase throughput, but
+	 * also increases the memory consumption. Must be a positive number and a power of 2.
+	 *
 	 * @param queueSize the queue size
 	 */
 	public void setQueueSize(int queueSize) {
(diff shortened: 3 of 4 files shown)


src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -289,8 +288,7 @@ public void setAutoStartup(boolean autoStartup) {
 		this.autoStartup = autoStartup;
 	}
 
-	@Override
-	public void stop(Runnable callback) {
+	@Override public void stop(Runnable callback) {
```
garyrussell: new IDEA?

Imports too.

garyrussell:
This is a pretty big change to have no tests :smile: 

Given the commit comment mentions tests, I guess this was due to the commit hour.

src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -590,38 +591,35 @@ public Long safeValueOf(Partition object) throws Exception {
 
 	}
 
-	@SuppressWarnings("serial")
-	private class PartitionToLeaderFunction implements Function<Partition, BrokerAddress> {
+	@SuppressWarnings("serial") private class PartitionToLeaderFunction implements Function<Partition, BrokerAddress> {
```
garyrussell: ditto

src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
mbogoevici: Oops.

mbogoevici:
Yeah, actually I've tested this manually, since tests that involve starting and stopping brokers are faring pretty badly with the PR. 
Let me add one though.

mbogoevici:
The test battery should catch any regressions though.

mbogoevici:
The 'test refactoring' was an artifact of an earlier change that I ditched on rebase, since it didn't really improve much over the existing test set.

src/main/java/org/springframework/integration/kafka/listener/ConcurrentMessageListenerDispatcher.java
```diff
@@ -120,19 +120,17 @@ public void dispatch(KafkaMessage message) {
...
+					queueSize, offsetManager, delegateListener, errorHandler, taskExecutor, autoCommitOnError);
 			delegateList.add(queueingMessageListenerInvoker);
 		}
 
 		// evenly distribute partitions across delegates
 		delegates = Maps.mutable.of();
-		int i = 0;
 		for (Partition partition : partitions) {
-			delegates.put(partition, delegateList.get((i++) % consumers));
+			delegates.put(partition, delegateList.get((partitionAssignmentCounter.incrementAndGet()) % consumers));
```
garyrussell: Is it ok that this is not reset to 0 on the next `start()` ?

mbogoevici: That'd probably be a better idea, although in practice this is just a round-robining counter.

src/main/java/org/springframework/integration/kafka/core/DefaultConnectionFactory.java
```diff
@@ -38,13 +23,25 @@
...
+import org.springframework.util.Assert;
 import scala.collection.JavaConversions;
 
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
```
artembilan: And again some mess with the import order...
Not a big deal, of course, but it is an extra noise during review...

src/main/java/org/springframework/integration/kafka/listener/ConcurrentMessageListenerDispatcher.java
garyrussell: Was there some other purpose of changing it from a local int to an `AtomicInteger` ?

src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -207,27 +203,29 @@ public Executor getFetchTaskExecutor() {
...
 	public Executor getAdminTaskExecutor() {
 		return adminTaskExecutor;
 	}
 
 	/**
-	 * The task executor for leader and offset updates
-	 * @param adminTaskExecutor the task executor for leader and offset updates.
+	 * The task executor for leader, offset, and partition reassignment updates
+	 *
+	 * @param adminTaskScheduler the task executor for leader, offset and partition reassignment updates
```
artembilan: No blank lines in the method-level JavaDocs.

src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -207,27 +203,29 @@ public Executor getFetchTaskExecutor() {
...
 	 */
-	public void setAdminTaskExecutor(Executor adminTaskExecutor) {
-		this.adminTaskExecutor = adminTaskExecutor;
+	public void setAdminTaskExecutor(Executor adminTaskScheduler) {
```
artembilan: Haven't you missed anything here?
Otherwise what is the reason to rename it to the `Scheduler`...

src/main/java/org/springframework/integration/kafka/listener/ConcurrentMessageListenerDispatcher.java
mbogoevici: We should revert this change altogether. It's an artifact of an earlier intent to support adding partitions dynamically, case in which it would allow to round robin over the set of listeners - ensuring that the partition distribution is balanced. Since we don't do that, it can revert to a local variable.

src/main/java/org/springframework/integration/kafka/core/DefaultConnectionFactory.java
mbogoevici: Oops. Will revert.

src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -600,15 +616,15 @@ public BrokerAddress valueOf(Partition partition) {
...
-	@SuppressWarnings("serial")
-	private class LaunchFetchTaskProcedure implements Procedure<BrokerAddress> {
-
-		@Override
-		public void value(BrokerAddress brokerAddress) {
-			fetchTaskExecutor.execute(new FetchTask(brokerAddress));
-		}
-
-	}
+	// @SuppressWarnings("serial")
```
mbogoevici: This needs to be removed.

src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
mbogoevici: No, it should be renamed back.

src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
mbogoevici: Also, dot at the end of the sentence.

src/test/java/org/springframework/integration/kafka/listener/NewBrokerRecoveryTests.java
```diff
@@ -0,0 +1,141 @@
...
+	public void testCompleteShutdown() throws Exception {
+		int partitionCount = 1;
+
+		createTopic(TEST_TOPIC, partitionCount, 2, 2);
+
+		// stop one Kafka instance - the partition is now underreplicated and only one
+		kafkaEmbeddedBrokerRule.bounce(0, false);
+
+		// sleep to let things settle
+		Thread.sleep(SETTLEMENT_TIME_IN_MILLIS);
```
artembilan: Yeah... This isn't good solution for our Bamboo...
Can't we iterate with less timeout and check something, e.g. using `TestUtils.getProperty()` ?

mbogoevici: I assume that's the result of the frustration with timeout tests in SI :).

Trying to think of a reliable 'something' to test for.

mbogoevici: Yeah, maybe there's a way to test this more reliably. I suppose we can try to see if the broker leaves the in-sync replica list.

src/test/java/org/springframework/integration/kafka/listener/NewBrokerRecoveryTests.java
```diff
@@ -0,0 +1,141 @@
...
+		// now start sending messages again
+		createMessageSender("none").send(createMessagesInRange(149, 199, TEST_TOPIC,partitionCount));
+
+
+		latch.await(50, TimeUnit.SECONDS);
+		kafkaMessageListenerContainer.stop();
+
+		assertThat(receivedData.valuesView().toList(), hasSize(expectedMessageCount));
+		assertThat(latch.getCount(), equalTo(0L));
+		System.out.println("All messages received ... checking ");
```
artembilan: Is that intentional ?
Should we remove it ?

src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -326,12 +325,20 @@ public void start() {
...
+				}
+				if (adminTaskExecutor == null) {
+					adminTaskExecutor = Executors.newSingleThreadExecutor();
+				}
+				for (Pair<BrokerAddress, RichIterable<Partition>> entry : partitionsByLeader.keyMultiValuePairsView()) {
+					FetchTask fetchTask = new FetchTask(entry.getOne(), entry.getTwo());
+					fetchTaskExecutor.execute(fetchTask);
+					fetchTasksByBroker.put(entry.getOne(), fetchTask);
 				}
-				partitionsByBrokerMap.forEachKey(launchFetchTask);
```
artembilan: It's out of the change here, but shouldn't we move `this.running = true;` to the end of the `start()` ?

artembilan: Another out of topic concern:
the `this.offsetManager` population here in the `start()`.
The `stop()` doesn't check it for `null` - just because we may not start it at all, but can call `stop()`. 
Something like regression testing :smile:

src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -18,23 +18,30 @@
 
 import static com.gs.collections.impl.utility.ArrayIterate.flatCollect;
 import static com.gs.collections.impl.utility.Iterate.partition;
-import static com.gs.collections.impl.utility.MapIterate.forEachKeyValue;
```
artembilan: No `static` imports in the production code.

src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
mbogoevici: What do you mean? `start()` and `stop()` are mutually exclusive via lifecycle monitor, and the code in `stop()` checks if the component is running.

artembilan: Sure! Never mind.
Anyway `setOffsetManager()` should have `Assert.notNull()`.

mbogoevici: Since the running tasks will terminate when `running` is `false`, we should set the flag before launching them. If the method throws an exception, the whole component will not be started anyway.

artembilan: Yes, I see:
```
try {
				while (active && isRunning()) {
............
finally {
				active = false;
				synchronized (fetchTasksByBroker) {
```
Never mind for now.

src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -372,89 +385,82 @@ public boolean isLongLived() {
...
 									}
 								}
+								// handle errors
+								if (result.getErrors().size() > 0) {
+									// find partitions with leader errors and
+									PartitionIterable<Map.Entry<Partition, Short>> partitionByLeaderErrors = partition(
+											result.getErrors().entrySet(), new IsLeaderErrorPredicate());
+									RichIterable<Partition> partitionsWithLeaderErrors = partitionByLeaderErrors
+											.getSelected().collect(keyFunction);
+									resetLeaders(partitionsWithLeaderErrors);
```
artembilan: I think we should `listenedPartitions.removeAllIterable(partitionsWithLeaderErrors)` before scheduling `resetLeaders`.
We may go to the next iteration here, but the `resetLeaders` won't be evaluated yet, so we are going to visit this error handling block one more time and for the same errors.

Looks like the reset/remove order does not make sense because we are `synchronized (listenedPartitions)` here and in the `addListenedPartitionsIfActive()`.

src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -372,89 +385,82 @@ public boolean isLongLived() {
...
+										fetchOffsets.replace(batch.getPartition(), highestFetchedOffset);
 									}
 								}
+								// handle errors
+								if (result.getErrors().size() > 0) {
+									// find partitions with leader errors and
+									PartitionIterable<Map.Entry<Partition, Short>> partitionByLeaderErrors = partition(
+											result.getErrors().entrySet(), new IsLeaderErrorPredicate());
+									RichIterable<Partition> partitionsWithLeaderErrors = partitionByLeaderErrors
+											.getSelected().collect(keyFunction);
```
artembilan: I don't see reason for extraction for that `keyFunction` property.
The inline `Functions.getKeyFunction()` usage looks good.
Although this is just my preference. 
It's difficult to understand this GS Collections `partition` code, especially when we deal with partitions domain as well :smile:

src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -372,89 +385,82 @@ public boolean isLongLived() {
...
+									resetLeaders(partitionsWithLeaderErrors);
+
+									PartitionIterable<Map.Entry<Partition, Short>> partitionsWithOffsetsOutOfRange = partitionByLeaderErrors
+											.getRejected().partition(new IsOffsetOutOfRangePredicate());
+									resetOffsets(
+											partitionsWithOffsetsOutOfRange.getSelected().collect(keyFunction).toSet());
+									// it's not a leader issue
+									listenedPartitions.removeAllIterable(
```
artembilan: I think remove **all** `partitionsWithLeaderErrors` should be enough.

src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -372,89 +385,82 @@ public boolean isLongLived() {
...
+							else {
+								active = false;
 							}
 						}
 						catch (ConsumerException e) {
-							resetLeaders(fetchPartitions.toImmutable());
+							active = false;
+							// the connection is broken, terminate the task
+							kafkaTemplate.getConnectionFactory().disconnect(brokerAddress);
+							resetLeaders(listenedPartitions.toImmutable());
```
artembilan: Since we are going to leave the loop and `fetchTasksByBroker.remove(brokerAddress)`  in the `finally`, I think it would be safer to `resetLeaders` after `fetchTasksByBroker.remove()` there in the `finally`, rather than here and finish with the unreachable broker.

src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
mbogoevici: I don't understand the comment. Isn't that what resetLeaders actually does? Remove the partitions and schedule the task?

src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -462,27 +468,18 @@ public void run() {
 			}
 		}
 
-
 		private void resetLeaders(final Iterable<Partition> partitionsToReset) {
-			stopFetchingFromPartitions(partitionsToReset);
+			listenedPartitions.removeAllIterable(partitionsToReset);
```
artembilan: Yeah... I see now.
Never mind.
Sorry for the noise above...

src/main/java/org/springframework/integration/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -503,10 +500,23 @@ public void run() {
...
+
+						MutableSetMultimap<BrokerAddress, Partition> partitionsByBroker = UnifiedMap
+								.newMap(kafkaTemplate.getConnectionFactory().getLeaders(partitionsToReset)).flip();
+						for (Pair<BrokerAddress, RichIterable<Partition>> pair : partitionsByBroker
+								.keyMultiValuePairsView()) {
+							synchronized (fetchTasksByBroker) {
+								boolean addedSuccessfully = false;
+								FetchTask fetchTask = fetchTasksByBroker.get(pair.getOne());
+								if (fetchTask != null) {
+									addedSuccessfully = fetchTask.addListenedPartitionsIfActive(pair.getTwo());
```
artembilan: Yeah... There is no the problem with the `resetLeaders` in `catch` above.
You check the result of `addListenedPartitionsIfActive` here and start a new fresh task replacing the old one.

Well, I'm giving up. So distributed logic breaks my mind :smile: 

Maybe tomorrow from the fresh morning...

src/test/java/org/springframework/integration/kafka/listener/NewBrokerRecoveryTests.java
```diff
@@ -0,0 +1,141 @@
...
+		ArrayList<Partition> readPartitions = new ArrayList<>();
+		readPartitions.add(new Partition(TEST_TOPIC, 0));
+		final KafkaMessageListenerContainer kafkaMessageListenerContainer =
+				new KafkaMessageListenerContainer(connectionFactory,
+						readPartitions.toArray(new Partition[readPartitions.size()]));
+		kafkaMessageListenerContainer.setMaxFetch(100);
+		kafkaMessageListenerContainer.setConcurrency(1);
+
+		final int expectedMessageCount = 200;
+		createMessageSender("none").send(createMessagesInRange(0, 49, TEST_TOPIC,partitionCount));
```
artembilan: Not sure how the `Sender` should work, but this test hangs for me here with:
```

File: src/test/java/org/springframework/integration/kafka/listener/NewBrokerRecoveryTests.java [hunk 19]
```diff
@@ -0,0 +1,137 @@
...
+import org.springframework.integration.kafka.core.Partition;
+import org.springframework.integration.kafka.rule.KafkaEmbedded;
+
+/**
+ * @author Marius Bogoevici
+ */
+
+public class NewBrokerRecoveryTests extends AbstractMessageListenerContainerTests {
+
+	public static final int SETTLEMENT_TIME_IN_MILLIS = 1000;
```
artembilan: So, this is redundant now.
I can fix it on merge, if my Windows will accept the fix :smile: 
On Mac it works good!

src/test/java/org/springframework/integration/kafka/rule/KafkaEmbedded.java
```diff
@@ -212,13 +251,44 @@ public void restart(final int index) throws Exception {
...
+			}
+			catch (InterruptedException e) {
+				break;
+			}
+			canExit = true;
+			TopicMetadata topicMetadata = AdminUtils$.MODULE$.fetchTopicMetadataFromZk(topic, getZkClient());
+			if (Errors.forCode(topicMetadata.errorCode()).exception() == null) {
+				for (PartitionMetadata partitionMetadata :
+						JavaConversions.asJavaCollection(topicMetadata.partitionsMetadata())) {
+					Collection<Broker> isr = JavaConversions.asJavaCollection(partitionMetadata.isr());
```
artembilan: `isr()`?
Perfect method name! Especially when you don't have source code... :angry:

mbogoevici: In-Sync Replica :) We could rename the variable though.

artembilan:
That's all for me here.
Will take a look one more time when will be at home this afternoon.
But everything looks good anyway. I don't see now critical concerns, unless it is still enough difficult to determine the whole logic when it is distributed between thread...
From other side I still don't feel good with those GC Collections :cry:

artembilan:
@mbogoevici , does it say anything to you?
```
15:31:03.233 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@0
15:31:03.237 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@2
15:31:03.239 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@4
15:31:03.241 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@6
15:31:03.242 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@8
15:31:03.243 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@10
15:31:03.244 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@12
15:31:03.245 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@14
15:31:03.246 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@16
15:31:03.248 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@18
15:31:03.249 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@20
15:31:03.251 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@22
15:31:03.252 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@24
15:31:03.253 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@26
15:31:03.254 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@28
15:31:03.257 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@30
15:31:03.258 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@32
15:31:03.260 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@34
15:31:03.261 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@36
15:31:03.262 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@38
15:31:03.263 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@40
15:31:03.265 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@42
15:31:03.267 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@44
15:31:03.268 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@46
15:31:03.270 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@48
15:31:03.374 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@50
15:31:03.476 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@50
15:31:03.579 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@50
15:31:03.683 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@50
15:31:03.785 DEBUG [kafka-fetch-1][org.springframework.integration.kafka.core.DefaultConnection] Reading from Partition[topic='test-topic', id=0]@50
```
and so on for the `50`

mbogoevici:
Well, that's what you get when debug is enabled for `[org.springframework.integration.kafka.core.DefaultConnection. It's disabled by default in the tests.

garyrussell:
I guess we could consider dropping it to trace level. @mbogoevici are we expecting a final commit to address the loose ends?

mbogoevici:
I guess https://github.com/spring-projects/spring-integration-kafka/pull/89#issuecomment-157423426, I guess we just have a couple of minor things left: https://github.com/spring-projects/spring-integration-kafka/pull/89#discussion_r45081740 and https://github.com/spring-projects/spring-integration-kafka/pull/89#discussion_r45081556. The main question there was running the tests in Windows.

garyrussell:
Oh, sorry; I see you reverted that AtomicInteger stuff - @artembilan are you going to merge after testing on Windoze?

PR #100, opened 2016-01-06
INTEXT-216: KafkaProducerContext.beanName as null

artembilan:
JIRA: https://jira.spring.io/browse/INTEXT-216

Having the default value for the `NamedComponent` contract (`beanName`)
 doesn't allow us to override (generate) the unique bean name for the component.

 Clean up `KafkaProducerContext.beanName` to `null` by default.

**Cherry-pick to 1.3.x**


artembilan:
I found it as so simple fix that I'm going to merge it myself.
To save some time.
Review and objections are welcome, of course. But yes: already on the commit in the master...

PR #120, opened 2016-05-31
Rename `reconnectInterval` to `recoveryInterval`

mbogoevici:


garyrussell:
Do we really want to make a breaking API change in a point release?

artembilan:
@garyrussell , it's not breaking change. Just see the previous commit in `master`. And pay attention to the `@since` on the setter.
That just was my fault do not think here about API alignment with Spring AMQP...

artembilan:
Sorry, not `master` - `1.3.x`

garyrussell:
Oops - got it; thanks 😄

mbogoevici:
Yeah, sorry - I should have made that clearer in the commit message :)

Issue #125, opened 2016-07-18
Reinstate the ability to send messages synchronously


mbogoevici:
In the 1.3 version we used to have a setting for the `KafkaProducingMessasgeHandler` that allowed opting between asynchronous (default) and synchronous sending of messages (via the old `ProducerConfiguration`). 

While asynchronous sending is preferred for performance reasons, synchronous sending allows the sender to block until a message is received and to ensure that any sending exception is thrown in the caller's thread

With the migration to the `KafkaTemplate`, the ability of sending messages synchronously seems to have been lost.

artembilan:
From big height I don't mind for such a feature.

Only concern is that I'd like to move this project into Spring Integraiton 5.0. 
Therefore I don't want to have any changes here until that time.

If that is only IMO and we can still live with this project in the separate repo, let's consider to fix it as it is proposed. 

But bear in mind that we can't expose XML `sync` attribute until `2.1` version.

Any other thoughts are welcome!

CC @sobychacko

mbogoevici:
Yeah, it's blocking a pretty important feature in the 1.1 Kafka binder, unfortunately.

PR #127, opened 2016-07-21
GH-125: Introduce `sync` mode for KafkaProducerMH

artembilan:
Fixes: GH-125 (https://github.com/spring-projects/spring-integration-kafka/issues/125)

* Add `sync` mode for the `KafkaProducerMessageHandler` (`false` by default) to wait for result from the send `Future`
* Add `sendTimeout` do not block `Future.get()` forever
* Provide test-case based on the `MockProducer` to complete the record with an exception and verify that the `sync` mode works well
* Some dependencies upgrade

### DIFF
--- a/src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java
+++ b/src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java
@@ -152,7 +152,7 @@ public class KafkaProducerMessageHandler<K, V> extends AbstractMessageHandler {
 					future.get(this.sendTimeout, TimeUnit.MILLISECONDS);
 				}
 				catch (TimeoutException te) {
-					throw new MessageTimeoutException(message,"Timeout waiting for response from KafkaProducer", te);
+					throw new MessageTimeoutException(message, "Timeout waiting for response from KafkaProducer", te);
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java
```diff
@@ -94,20 +123,30 @@ protected void handleMessageInternal(final Message<?> message) throws Exception
...
+				future = this.kafkaTemplate.send(topic, partitionId, (K) messageKey, (V) message.getPayload());
+			}
+		}
+		if (this.sync) {
+			if (this.sendTimeout < 0) {
+				future.get();
 			}
 			else {
-				this.kafkaTemplate.send(topic, partitionId, (K) messageKey, (V) message.getPayload());
+				future.get(this.sendTimeout, TimeUnit.MILLISECONDS);
```
garyrussell: Perhaps we should wrap the `TimeoutException` in a `MessageTimeoutException` with `failedMessage` set to `message`. Otherwise we just get a generic `MessageHandlingException`.

mbogoevici:
Squashed and merged https://github.com/spring-projects/spring-integration-kafka/commit/7cc94a9fa2656908dbe8b04649b7736ed2c0e365

Issue #158, opened 2017-02-22
Add support for record timestamp


artembilan:
Since Spring Kafka now supports [record timestamp](https://github.com/spring-projects/spring-kafka/issues/227), it sounds organic to provide its support on the Spring Integration level as well.

The `KafkaMessageDrivenChannelAdapter` might not need any changes because of `MessagingMessageConverter` changes on the Spring Kafka side, but `KafkaProducerMessageHandler` needs an addition to resolve a `timestamp` at runtime against request message. Fully similar what we have for `key`, `topic` and `partition` already. The `KafkaProducerMessageHandlerSpec` must be fixed respectively.

bijukunjummen:
Working on it now @artembilan , because of API changes in `spring-kafka` looks like this project is broken, will work on fixing it first.

artembilan:
Ah! Sure. We changes `KafkaOperations` to avoid types clashing.

Thank you, @bijukunjummen, for taking care of it!

PR #159, opened 2017-02-22
Fixing build - replaced a call to a removed API in spring-kafka project

bijukunjummen:
@artembilan, this is just to fix the build, may be this can be merged in first and I can get the timestamp addition in a separate PR ?

Related to #158


bijukunjummen:
No longer required..a new PR #160 covers this.

PR #160, opened 2017-02-23
GH-158 - Support for timestamps in inbound and outbound messages

bijukunjummen:
#158 - Support for timestamps in inbound and outbound messages

GH Issue Link - https://github.com/spring-projects/spring-integration-kafka/issues/158

Upstream Spring-Kafka issue - https://github.com/spring-projects/spring-kafka/issues/227
Upstream Spring-Kafka PR - https://github.com/spring-projects/spring-kafka/pull/241

* Added a `timestamp-expression` to outbound channel adapter schema
* Added support for specifying a timestamp expression to Kafka Java DSL

### DIFF
--- a/src/main/java/org/springframework/integration/kafka/dsl/KafkaProducerMessageHandlerSpec.java
+++ b/src/main/java/org/springframework/integration/kafka/dsl/KafkaProducerMessageHandlerSpec.java
@@ -195,6 +195,51 @@ public class KafkaProducerMessageHandlerSpec<K, V>
 	}
 
 	/**
+	 * Configure a SpEL expression to determine the timestamp at runtime against a
+	 * request Message as a root object of evaluation context.
+	 *
+	 * @param timestampExpression the timestamp expression to use.
+	 * @return the spec.
+	 *
+	 * @since 3.0
+	 */
+	public KafkaProducerMessageHandlerSpec<K, V> timestampExpression(String timestampExpression) {
+		return this.timestampExpression(PARSER.parseExpression(timestampExpression));
+	}
+
+	/**
+	 * Configure a {@link Function} that will be invoked at run time to determine the Kafka record timestamp
+	 * will be stored in the topic. Typically used with a Java 8 Lambda expression:
+	 * <pre class="code">
+	 * {@code
+	 * .timestamp(m -> m.getHeaders().get("mytimestamp_header", Long.class))
+	 * }
+	 * </pre>
+	 * @param timestampFunction the partitionId function.
+	 * @param <P> the expected payload type.
+	 * @return the spec.
+	 *
+	 * @since 3.0
+	 */
+	public <P> KafkaProducerMessageHandlerSpec<K, V> timestamp(Function<Message<P>, Long> timestampFunction) {
+		return timestampExpression(new FunctionExpression<>(timestampFunction));
+	}
+
+	/**
+	 * Configure an {@link Expression} to determine the timestamp at runtime against a
+	 * request Message as a root object of evaluation context.
+	 * @param timestampExpression the timestamp expression to use.
+	 * @return the spec.
+	 *
+	 * @since 3.0
+	 */
+	public KafkaProducerMessageHandlerSpec<K, V> timestampExpression(Expression timestampExpression) {
+		this.target.setTimestampExpression(timestampExpression);
+		return _this();
+	}
+
+
+	/**
 	 * A {@code boolean} indicating if the {@link KafkaProducerMessageHandler}
 	 * should wait for the send operation results or not. Defaults to {@code false}.
 	 * In {@code sync} mode a downstream send operation exception will be re-thrown.
--- a/src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java
+++ b/src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2016 the original author or authors.
+ * Copyright 2013-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -43,6 +43,8 @@ import org.springframework.util.concurrent.ListenableFuture;
  * @author Artem Bilan
  * @author Gary Russell
  * @author Marius Bogoevici
+ * @author Biju Kunjummen
+ *
  * @since 0.5
  */
 public class KafkaProducerMessageHandler<K, V> extends AbstractMessageHandler {
@@ -142,28 +144,15 @@ public class KafkaProducerMessageHandler<K, V> extends AbstractMessageHandler {
 				? this.messageKeyExpression.getValue(this.evaluationContext, message)
 				: message.getHeaders().get(KafkaHeaders.MESSAGE_KEY);
 
-		ListenableFuture<?> future;
+		Long timestamp = message.getHeaders().get(KafkaHeaders.TIMESTAMP, Long.class);
 
 		V payload = (V) message.getPayload();
 		if (payload instanceof KafkaNull) {
 			payload = null;
 		}
-		if (partitionId == null) {
-			if (messageKey == null) {
-				future = this.kafkaTemplate.send(topic, payload);
-			}
-			else {
-				future = this.kafkaTemplate.send(topic, (K) messageKey, payload);
-			}
-		}
-		else {
-			if (messageKey == null) {
-				future = this.kafkaTemplate.send(topic, partitionId, payload);
-			}
-			else {
-				future = this.kafkaTemplate.send(topic, partitionId, (K) messageKey, payload);
-			}
-		}
+
+		ListenableFuture<?> future = this.kafkaTemplate.send(topic, partitionId, timestamp, (K) messageKey, payload);
+
 		if (this.sync) {
 			Long sendTimeout = this.sendTimeoutExpression.getValue(this.evaluationContext, message, Long.class);
 			if (sendTimeout == null || sendTimeout < 0) {
--- a/src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java
+++ b/src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java
@@ -61,6 +61,8 @@ public class KafkaProducerMessageHandler<K, V> extends AbstractMessageHandler {
 
 	private volatile Expression partitionIdExpression;
 
+	private volatile Expression timestampExpression;
+
 	private boolean sync;
 
 	private Expression sendTimeoutExpression = new ValueExpression<>(DEFAULT_SEND_TIMEOUT);
@@ -80,6 +82,18 @@ public class KafkaProducerMessageHandler<K, V> extends AbstractMessageHandler {
 
 	public void setPartitionIdExpression(Expression partitionIdExpression) {
 		this.partitionIdExpression = partitionIdExpression;
+	}
+
+	/**
+	 * Specify a SpEL expression to evaluate a timestamp that will be added in the Kafka record.
+	 * The resulting value should be a {@link Long} type representing epoch time in milliseconds.
+	 *
+	 * @param timestampExpression the {@link Expression} for timestamp to wait for result
+	 * fo send operation.
+	 * @since 3.0.0
+	 */
+	public void setTimestampExpression(Expression timestampExpression) {
+		this.timestampExpression = timestampExpression;
 	}
 
 	public KafkaTemplate<?, ?> getKafkaTemplate() {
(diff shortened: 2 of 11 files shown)


src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java
```diff
@@ -142,28 +144,15 @@ protected void handleMessageInternal(final Message<?> message) throws Exception
 				? this.messageKeyExpression.getValue(this.evaluationContext, message)
 				: message.getHeaders().get(KafkaHeaders.MESSAGE_KEY);
 
-		ListenableFuture<?> future;
+		Long timestamp = message.getHeaders().get(KafkaHeaders.TIMESTAMP, Long.class);
```
artembilan: Everything is good, @bijukunjummen, and we can accept PR, but then we need to start a new story for improvements I will request. So, let's address then here, too! 😄 

So, this `KafkaHeaders.TIMESTAMP` is good, but I really would like to see `timestampExpression` as well.
Absolutely similar what we have for `topic `, `partitionId` and `messageKey` here already.

The new expression option should be reflected in the `KafkaProducerMessageHandlerSpec` for Java DSL.

Of course, if you have some argument do not do that, I'd like to hear it.

Also, please, take a look into `README.adoc`. Maybe there is need to mention our new `timestamp` support as well.

And one more general concern from my side: commit message - https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.adoc#provide-a-link-to-the-jira-issue-in-the-associated-pull-request

We prefer:
* to have a GH issue number in the header of the message
* the header must be not longer than 50 symbols
* After header there must be a blank line
* The next line should be a link to the GH issue (JIRA) to have a cross-link traceability when you read commits history
* When we talk about GH issues, there are some rules to  have them automatically closed when we apply commit to `upstream`: https://help.github.com/articles/closing-issues-via-commit-messages/

Something like that. Just to make our life easier. 
Otherwise I forgot to polish your commit message and we don't know now for which story is a commit in master: https://github.com/spring-projects/spring-kafka/commit/8f43ad21335650a119b82061b08c7fab24ca34cc 😄

bijukunjummen: Awesome, will make these changes and get back @artembilan , yes I was thinking about the expression part but was not sure. It does make sense to have a timeExpression, will introduce it and see how it goes.

src/main/java/org/springframework/integration/kafka/dsl/KafkaProducerMessageHandlerSpec.java
```diff
@@ -195,6 +195,29 @@
...
+		return this.timestampExpression(PARSER.parseExpression(timestampExpression));
+	}
+
+	/**
+	 * Configure an {@link Expression} to determine the timestamp at runtime against a
+	 * request Message as a root object of evaluation context.
+	 * @param timestampExpression the timestamp expression to use.
+	 * @return the spec.
+	 */
+	public KafkaProducerMessageHandlerSpec<K, V> timestampExpression(Expression timestampExpression) {
```
artembilan: Please, add also a `Function` variant.
See `partitionId(Function<Message<P>, Integer> partitionIdFunction)` for sample.

Also new API must be marked with `@since`: https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.adoc#follow-the-code-style
Just in case you would like to to avoid similar comments in the future contributions and let us save some time during merge 😉

src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java
bijukunjummen: @artembilan, I have updated the PR now. I have very likely missed some of your points, if you can have a quick review and let me know how it looks now.

src/test/java/org/springframework/integration/kafka/config/xml/KafkaOutboundAdapterParserTests.java
```diff
@@ -60,23 +61,27 @@
...
-		assertThat(messageHandler).isNotNull();
-		assertThat(messageHandler.getOrder()).isEqualTo(3);
-		assertThat(TestUtils.getPropertyValue(messageHandler, "topicExpression.literalValue")).isEqualTo("foo");
-		assertThat(TestUtils.getPropertyValue(messageHandler, "messageKeyExpression.expression")).isEqualTo("'bar'");
-		assertThat(TestUtils.getPropertyValue(messageHandler, "partitionIdExpression.expression")).isEqualTo("'2'");
-		assertThat(TestUtils.getPropertyValue(messageHandler, "sync", Boolean.class)).isTrue();
-		assertThat(TestUtils.getPropertyValue(messageHandler, "sendTimeoutExpression.expression")).isEqualTo("1000");
-
-		messageHandler
+		assertThat(messageHandler1).isNotNull();
```
artembilan: Please, explain what happened here? 
Why these lines looks the same but marked as modification?

src/test/java/org/springframework/integration/kafka/dsl/KafkaDslTests.java
```diff
@@ -178,15 +181,18 @@ public IntegrationFlow sendToKafkaFlow() {
...
-							.subscribe(sf -> sf.handle(kafkaMessageHandler(producerFactory(), TEST_TOPIC),
-									e -> e.id("kafkaProducer")))
-							.subscribe(sf -> sf.handle(kafkaMessageHandler(producerFactory(), TEST_TOPIC3),
-									e -> e.id("kafkaProducer3")))
+							.subscribe(sf -> sf.handle(
+									kafkaMessageHandler(producerFactory(), TEST_TOPIC1)
+											.timestampExpression("T(Long).valueOf('1487694048633')"),
+									e -> e.id("kafkaProducer1")))
+							.subscribe(sf -> sf.handle(kafkaMessageHandler(producerFactory(), TEST_TOPIC2)
+											.timestampExpression("T(System).currentTimeMillis()"),
```
artembilan: When you have a `Function` variant, this will be pretty straightforward for Java codding:
```
.timestampExpression(m -> System.currentTimeMillis())
```

src/test/java/org/springframework/integration/kafka/dsl/KafkaDslTests.java
```diff
@@ -178,15 +181,18 @@ public IntegrationFlow sendToKafkaFlow() {
 			return f -> f
 					.<String>split(p -> Stream.generate(() -> p).limit(101).iterator(), null)
 					.publishSubscribeChannel(c -> c
-							.subscribe(sf -> sf.handle(kafkaMessageHandler(producerFactory(), TEST_TOPIC),
-									e -> e.id("kafkaProducer")))
-							.subscribe(sf -> sf.handle(kafkaMessageHandler(producerFactory(), TEST_TOPIC3),
-									e -> e.id("kafkaProducer3")))
+							.subscribe(sf -> sf.handle(
+									kafkaMessageHandler(producerFactory(), TEST_TOPIC1)
+											.timestampExpression("T(Long).valueOf('1487694048633')"),
```
artembilan: FYI. The SpEL is pretty smart with type conversions.
So, there is no reason to explicitly say which type for string you want. The `ConversionService` just easily will use `Long.valueOf()`  for your simple expression like : `.timestampExpression("'1487694048633'")`.
See `StringToNumberConverterFactory`.

src/test/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandlerTests.java
```diff
@@ -120,4 +122,65 @@ public void testOutbound() {
...
+		handler.setBeanFactory(mock(BeanFactory.class));
+		handler.afterPropertiesSet();
+
+		Message<?> message = MessageBuilder.withPayload("foo")
+				.setHeader(KafkaHeaders.TOPIC, topic1)
+				.setHeader(KafkaHeaders.MESSAGE_KEY, 2)
+				.setHeader(KafkaHeaders.PARTITION_ID, 1)
+				.build();
+
+		handler.setTimestampExpression(new SpelExpressionParser().parseExpression("T(Long).valueOf('1487694048633')"));
```
artembilan: You can use here simple `new ValueExpression<>(1487694048633)`.

src/test/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandlerTests.java
```diff
@@ -120,4 +122,65 @@ public void testOutbound() {
...
+		handler.handleMessage(message);
+
+		ConsumerRecord<Integer, String> record1 = KafkaTestUtils.getSingleRecord(consumer, topic1);
+		assertThat(record1).has(key(2));
+		assertThat(record1).has(partition(1));
+		assertThat(record1).has(value("foo"));
+		assertThat(record1).has(timestamp(1487694048633L));
+
+		Long currentTimeMarker = System.currentTimeMillis();
+		handler.setTimestampExpression(new SpelExpressionParser().parseExpression("T(System).currentTimeMillis()"));
```
artembilan: And this one can be replaced with the `new FunctionExpression<Message<?>>(m -> System.currentTimeMillis())`.
Not so critical, just in case to study something new yet 😄

src/test/java/org/springframework/integration/kafka/config/xml/KafkaOutboundAdapterParserTests.java
bijukunjummen: Ha, not sure :-), will check, may have done something and reverted

bijukunjummen: Ah, I remembered after looking at the code. I introduced two local variables `messageHandler1` and `messageHandler2`, this way the assignments and assertions to them remain separate.

src/test/java/org/springframework/integration/kafka/dsl/KafkaDslTests.java
bijukunjummen: Very neat, did not know that. I think I will keep the `Long.valueOf` just to show that a reasonably complex expression is possible

src/test/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandlerTests.java
bijukunjummen: Awesome work with DSL's @artembilan, a lot to take in, I understand parts of how the DSL is put together but there is a lot more to absorb :-)

artembilan: ??? Not sure in your latest " there is a lot more to absorb". Please elaborate more.

BTW, haven't you just pushed with `amend commit`? 
Please, don't do that during PR review. Just push new changes! That way the next review round might require only taking a look into that new commit

src/test/java/org/springframework/integration/kafka/config/xml/KafkaOutboundAdapterParserTests.java
artembilan: Doh. I have missed that part.
Just a bad habit to review only from GH 😄

src/test/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandlerTests.java
bijukunjummen: Ah, okay, yes makes sense, I have been squashing before committing. Will not do that in future. "lot more to absorb" from a learning perspective :-)

src/main/java/org/springframework/integration/kafka/dsl/KafkaProducerMessageHandlerSpec.java
```diff
@@ -195,6 +195,51 @@
...
+	 * <pre class="code">
+	 * {@code
+	 * .timestamp(m -> m.getHeaders().get("mytimestamp_header", Long.class))
+	 * }
+	 * </pre>
+	 * @param timestampFunction the partitionId function.
+	 * @param <P> the expected payload type.
+	 * @return the spec.
+	 *
+	 * @since 3.0
```
artembilan: Doh. We don't need `@since` here. This class has been fully introduced exactly `@since 3.0` 😄

src/test/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandlerTests.java
artembilan: 😄 Well, you come come to SI core and contribute some missed Java DSL components. 
E.g. there is a start for `MongoDb`, but I'm not sure that it would be enough in the end for GA...

src/test/java/org/springframework/integration/kafka/config/xml/KafkaOutboundAdapterParserTests.java
artembilan: Well, I think we don't need that extra variable because we don't use the first instance after obtaining the second.
So, replacing it is a way to go

artembilan:
Merging BTW.
And doing all those minor polishing there already...

artembilan:
Merged as https://github.com/spring-projects/spring-integration-kafka/commit/3a15302a77c2c90280660bd12b5ae7915b775219.

One more awesome contribution, @bijukunjummen .

Looking forward for more! 😄

bijukunjummen:
Awesome, thanks for your detailed review @artembilan, learned a lot from your comments.

PR #167, opened 2017-05-24
Adapt Java DSL to the latest SI

artembilan:
* Add ability to specify `id` for `MessageListenerContainer` and
`KafkaTemplate` beans configured internally and exposed as beans
by the DSL


garyrussell:
Fixed checkstyle and merged as 78591faca937806cc62daaefee42dfa9a899911a

PR #193, opened 2018-01-30
Polish synchronization in KafkaMessageSource

marknorkin:
Commit summary:
- As `running` is always accessed through KafkaMessageSource monitor use existing synchronized to guard access to it.
Also move setting `running=true` from consumerMonitor in createConsumer method to doReceive as it's guarded by needed lock
- Extract common stopping consumer logic


artembilan:
Cool! 

Thank you very much for the contribution.

Looking forward for more!

Issue #212, opened 2018-09-05
Add seek management support for the Kafka Inbound Channel Adapters


alpert:
Hi. 

This may be a generic question about Kafka and not directly related to spring integration but without manual offset management what will happen if i close and open my application(that using spring integration kafka) again? Will it continue where it left? Will it continue from beginning or end? What should I do if i want it to read from beginning, end of where it left?

artembilan:
The Spring Integration Kafka is fully based now on the Spring for Apache Kafka: https://spring.io/projects/spring-kafka

The `KafkaMessageDrivenChannelAdapter` and `KafkaInboundGateway` fully rely on the `MessageListenerContainer` therefore all the offset management comes from there: https://docs.spring.io/spring-kafka/docs/2.1.9.RELEASE/reference/html/_reference.html#committing-offsets.
If you use a `enable.auto.commit = true` , then you should understand that Kafka Client commits offsets in async manner, therefore it is indeed that after closing app and restarting you may start to consume from some earlier offset.

To re-read from the beginning you need to use a `ConsumerSeekCallback`: https://docs.spring.io/spring-kafka/docs/2.1.9.RELEASE/reference/html/_reference.html#seek.

Well, it is used, but looks like in case of  `KafkaMessageDrivenChannelAdapter` and `KafkaInboundGateway` we don't provide any hooks for that functionality.
And from here it sounds like we need to provide for those Channel Adapter a functionality like `setConsumerSeekAware()`. So, you will be able to do the same what can be done now with the `@KafkaListener` and `ConsumerSeekAware` impl on that class.

Saying that I'd recommend you as a workaround to use a `@KafkaListener`  for time being and send messages to the channel from the method impl or delegate to some `@MessagingGateway`.

Does it make sense?

@garyrussell , WDYT?

garyrussell:
Yes; we should add seek capability to the adapter.

harshita-mangly-mohan:
Hi, @garyrussell @artembilan,

I am using ConcurrentMessageListenerContainer with KafkaMessageDrivenChannelAdapter for an application Kafka implementation, in a use case to replay older messages from a kafka topic I plan to follow ->
Find offset by timestamp, seek and then continue polling from that offset

[...]

PR #213, opened 2018-09-05
GH-212: Add ConsumerSeekAware impl to Inbounds

artembilan:
Fixes spring-projects/spring-integration-kafka#212

**Cherry-pick to 3.0.x**


artembilan:
Do not merge yet: the proposed solution is slightly useless.
I believe we need to introduce a new header for the message to send. This is exactly what is a custom listener code - Spring Integration flow.

Stay tuned!

artembilan:
Ready for review.
Thanks

src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
```diff
@@ -205,13 +261,62 @@ protected AttributeAccessor getErrorMessageAttributes(Message<?> message) {
...
+		if (this.setAdditionalHeaders) {
+			if (this.messageListenerContainer instanceof ConcurrentMessageListenerContainer) {
+
+				seekCallbackHolder.set(callback);
+			}
+			else {
+				this.seekCallback = callback;
+			}
+		}
+	}
```
garyrussell: Not sure we need this extra complexity - why not always use the ThreadLocal?

src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
```diff
@@ -133,6 +150,45 @@ public void setRecoveryCallback(RecoveryCallback<? extends Object> recoveryCallb
...
+		this.onIdleSeekCallback = onIdleSeekCallback;
+	}
+
+	/**
+	 * Set a @code boolean} flag to indicate that request message should have extra headers.
+	 * @param setAdditionalHeaders {@code boolean} flag to add or not extra headers into the message to send.
+	 * @since 3.0.4
+	 * @see IntegrationKafkaHeaders
+	 */
+	public void setAdditionalRequestHeaders(boolean setAdditionalHeaders) {
```
garyrussell: This name seems too general; it is specifically related to adding the seek callback to the headers.

src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
artembilan: Well, it's not end-user API, so I don't see reason to worry about complexity in our code since with the single-threaded container we will avoid `ThreadLocal` overhead altogether.

src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
artembilan: Yes, for now. I'm open for better name.
Or do you think it would be better to add a `boolean` setter for any new headers which may come in the future?

garyrussell:
Did you see my general review comment?

artembilan:
Good point about `ListenerContainerIdleEvent`! 
I will remove the `idleCallback` option.

What else should I pay attention in your review?
Looks like I've answered to all comments.

Thanks

garyrussell:
I mean I don't think we need to add the callback as a header, since the Consumer is already there as one.

artembilan:
Huh? So, we don't need `registerSeekCallback(ConsumerSeekCallback callback)` now at all?

I'll fix all your concern. And having the argument about `Consumer` header, I don't see a value in the `ThreadLocal` discussion. That code is just going to disappear!

garyrussell:
Indeed 😄

src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
```diff
@@ -212,6 +233,13 @@ protected AttributeAccessor getErrorMessageAttributes(Message<?> message) {
 			super(null, null);
 		}
 
+		@Override
+		public void onPartitionsAssigned(Map<TopicPartition, Long> assignments, ConsumerSeekCallback callback) {
+			if (KafkaInboundGateway.this.onPartitionsAssignedSeekCallback != null) {
+				KafkaInboundGateway.this.onPartitionsAssignedSeekCallback.accept(assignments, callback);
+			}
+		}
+
```
garyrussell: Why only from the record listener?

src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java
```diff
@@ -225,6 +231,20 @@ public void setPayloadType(Class<?> payloadType) {
 		this.batchListener.setFallbackType(payloadType);
 	}
 
+	/**
+	 * Specify a {@link BiConsumer} for seeks management during
+	 * {@link ConsumerSeekAware.ConsumerSeekCallback#onPartitionsAssigned(Map, ConsumerSeekAware.ConsumerSeekCallback)}
+	 * call from the {@link org.springframework.kafka.listener.KafkaMessageListenerContainer}.
+	 * This is called from the internal {@link RecordMessagingMessageListenerAdapter} implementation.
```
garyrussell: In this case, it's called from both.

src/main/java/org/springframework/integration/kafka/inbound/KafkaInboundGateway.java
artembilan: Huh? There is no `IntegrationBatchMessageListener` in the gateway.

Sorry, I'm not sure how to proceed.

artembilan: On other thought the batch processing does not make sense in the request-reply scenarios.
So, that's fully fine that we don't support `batch` there.

Please, elaborate with your concern.

Thanks

PR #266, opened 2019-04-17
GH-265: Ensure assignTimeout is used when no partitions are assigned

### LINKED ISSUE #265: Set assigned = false in KafkaMessageSource when partitions are revoked
When creating a new consumer, `KafkaMessageSource` subscribes to partition assigned events and sets `KafkaMessageSource.this.assigned` to `true`.

https://github.com/spring-projects/spring-integration-kafka/blob/8f08abe265259da67ef3517ca107c6eb65033207/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageSource.java#L402-L416

However, `KafkaMessageSource.this.assigned` is not toggled back to `false` when partitions are revoked. 

https://github.com/spring-projects/spring-integration-kafka/blob/8f08abe265259da67ef3517ca107c6eb65033207/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageSource.java#L387-L400

Since, the value of `assigned` is used to determine whether to use `pollTimeout` or `assignTimeout`, if the last event on a poll was partitions revoked then on the next poll `assignTimeout` should be used but it won't be used.

Walliee:
* Refactored to rely on assignedPartitions for determining whether to
use pollTimeout or assignTimeout
* Removed redundant assigned flag

fixes #265

### DIFF
--- a/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageSource.java
+++ b/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageSource.java
@@ -81,6 +81,7 @@ import org.springframework.util.Assert;
  * @author Gary Russell
  * @author Mark Norkin
  * @author Artem Bilan
+ * @author Anshul Mehra
  *
  * @since 3.0.1
  *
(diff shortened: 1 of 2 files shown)


garyrussell:
@Walliee Thanks for the contribution!

In future, please limit commit headlines to 50 chars max; thanks.

Cherry-picked to 3.1.x as 4ec64590f60e814fd80ec2eb16e570a32ea981be

Issue #272, opened 2019-06-20
Consider adding `payloadType` to KafkaMessageDrivenChannelAdapterSpec


Cameronjmayfield:
Unless I'm missing a simpler way to configure this, payloadType on the KafkaMessageDrivenChannelAdapterSpec would allow us to use a type aware message converter.

artembilan:
I think this is just an omission during introduction of `KafkaMessageDrivenChannelAdapter.setPayloadType()` in version `2.1.1`.

So, good catch! 
Feel free to contribute the fix!

I see you have some commit in your local copy already 😉

garyrussell:
Perhaps a separate issue, but we should change `payloadType` to `Type` so generics are retained for `ParameterizedType`, which can be obtained by reflection or via a `ParameterizedTypeReference`.

We might not be able to backport that though.

artembilan:
Well, I think we need to keep it as a `Class` because we can't simply provide a `Class` then from the XML config: the string value isn't converted to class properly.
In the latest Spring Integration components I have a pair of properties on the matter:
```
/**
	 * Specify the type of payload to be generated when the inbound HTTP request
	 * content is read by the converters/encoders.
	 * By default this value is null which means at runtime any "text" Content-Type will
	 * result in String while all others default to <code>byte[].class</code>.
	 * @param requestPayloadType The payload type.
	 */
	public void setRequestPayloadTypeClass(Class<?> requestPayloadType) {
		setRequestPayloadType(ResolvableType.forClass(requestPayloadType));
	}

	/**
	 * Specify the type of payload to be generated when the inbound HTTP request
	 * content is read by the converters/encoders.
	 * By default this value is null which means at runtime any "text" Content-Type will
	 * result in String while all others default to <code>byte[].class</code>.
	 * @param requestPayloadType The payload type.
	 */
	public void setRequestPayloadType(ResolvableType requestPayloadType) {
		this.requestPayloadType = requestPayloadType;
	}
```

The `ResolvableType` has this factory method:
```
/**
	 * Return a {@link ResolvableType} for the specified {@link ParameterizedTypeReference}.
	 * <p>Note: The resulting {@link ResolvableType} instance may not be {@link Serializable}.
	 * @param typeReference the reference to obtain the source type from
	 * @return a {@link ResolvableType} for the specified {@link ParameterizedTypeReference}
	 * @since 4.3.12
	 * @see #forType(Type)
	 */
[...]

PR #273, opened 2019-06-21
GH-272: Add payloadType to KafkaMessageDrivenChannelAdapterSpec.java

Cameronjmayfield:


src/main/java/org/springframework/integration/kafka/dsl/KafkaMessageDrivenChannelAdapterSpec.java
```diff
@@ -141,6 +142,17 @@ public S recoveryCallback(RecoveryCallback<? extends Object> recoveryCallback) {
 		return _this();
 	}
 
+	/**
+	 * When using a type-aware message converter (such as {@code StringJsonMessageConverter},
+	 * set the payload type the converter should create. Defaults to {@link Object}.
+	 * @param payloadType the type.
+	 * @return the spec
+	 */
+	public S payloadType(Class<V> payloadType) {
```
artembilan: No, this is not about `V`. This is a fallback type to which that `V` is going to be converted.
Well, at runtime it really doesn't matter because this generic type is going to be erased anyway.
But for the proper code logic we need to be explicit what we are going to do here.

Please, add also `@since` to this new public API.

Also would be great to have some test-case to cover such a functionality.

Cameronjmayfield: I took a guess on the version number. Is 3.2.0 correct?

artembilan: Correct. Thank you for reminding about that! 😉 
It is really hard to keep in mind all the versions we support in all our projects.

Well, we always can find the current version in the `gradle.properties`.

Cameronjmayfield: I've pushed a change with the version number.

src/main/java/org/springframework/integration/kafka/dsl/KafkaMessageDrivenChannelAdapterSpec.java
```diff
@@ -141,6 +142,18 @@ public S recoveryCallback(RecoveryCallback<? extends Object> recoveryCallback) {
...
 
+	/**
+	 * When using a type-aware message converter (such as {@code StringJsonMessageConverter},
+	 * set the payload type the converter should create. Defaults to {@link Object}.
+	 * @param payloadType the type.
+	 * @return the spec
+	 * @since 3.2.0
+	 */
+	public S payloadType(Class payloadType) {
+		this.target.setPayloadType(payloadType);
```
artembilan: Thanks. 

any chance to have a simple test-case to cover this change?

Cameronjmayfield: I split another json test that would set the payload type after the dsl. The new test sets the payload type in the dsl and has the same assertions the previous test did in its second half. Let me know if this is what you were looking for.

PR #281, opened 2019-08-17
GH-280: Add topicPattern support to KafkaMessageSource

### LINKED ISSUE #280: Add topicPattern support to KafkaMessageSource
It'd be good to support pattern-based subscriptions in `KafkaMessageSource` just like `KafkaMessageListenerContainer`

Walliee:
* Refactored to topics/topicPattern to be a property to
prevent constructor telescoping

resolves #280

### DIFF
--- a/src/main/java/org/springframework/integration/kafka/config/xml/KafkaInboundChannelAdapterParser.java
+++ b/src/main/java/org/springframework/integration/kafka/config/xml/KafkaInboundChannelAdapterParser.java
@@ -30,6 +30,7 @@ import org.springframework.util.StringUtils;
  * Parser for the inbound channel adapter.
  *
  * @author Gary Russell
+ * @author Anshul Mehra
  * @since 3.2
  *
  */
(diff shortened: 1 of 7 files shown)


artembilan:
@Walliee , 

thank you for the contribution, first of all!

So, before we jump into deep review, a general question: what is wrong with extra constructor(s) for `Pattern`? This is really a required option and our policy to make API as much clearer for end-users as possible. So, when you create a `KafkaMessageSource`, you definitely don't miss to specify or `topics` or `topicPattern`.

@garyrussell , 

What made us in the beginning do not support `topicPattern` subscription?

garyrussell:
>What made us in the beginning do not support topicPattern subscription?

Simply an omission.

If we add support for `Pattern`, we should also add support for manual partition assignment, via `TopicPartitionOffset` as well, but perhaps that's a different PR.

And, I agree about the CTOR.

artembilan:
> manual partition assignment


Yeah... One day in Spring Kafka we realized that we have too many ctors, so a compromise has been done in favor of `ContainerProperties` abstraction. Sounds like we are coming to the same direction here, but now we need another similar abstraction since the name is wrong for `MessageSource` implementation. Plus there are many other options which don't belong to polling solution... 🤔

garyrussell:
Maybe a new super class of `ContainerProperties` with common `ConsumerProperties` ?

artembilan:
I've raised an issue is Spring Kafka to consider that compromise instead of making setters in the `MessageSource`.

Thanks

Walliee:
> Maybe a new super class of ContainerProperties with common ConsumerProperties ?

I like it too but yeah only motive was to avoid explosion in number of constructors. 

I can update the PR once [spring-projects/spring-kafka#1203](https://github.com/spring-projects/spring-kafka/issues/1203) is done.

artembilan:
Hi, @Walliee !

Thank you for your feedback, but you know you may not wait for that issue to be resolved, but just contribute the fix for it yourself!

Walliee:
@artembilan Yes, happy to but I'll probably only be able to do it later today. I see @garyrussell already assigned it to himself so wonder if he's going to get to it before I do.

artembilan:
Well, actually there is no rush until September 5th: https://github.com/spring-projects/spring-integration-kafka/milestone/26. So, I think @garyrussell is OK, too, to wait for your contribution until then.
Thank you again!

garyrussell:
I am about to push a PR for 1203

garyrussell:
I also opened this https://github.com/spring-projects/spring-integration-kafka/issues/282

artembilan:
So, https://github.com/spring-projects/spring-kafka/issues/1203 has been fixed. We are going to have a BUILD-SNAPSHOT shortly, therefore you are good to poll latest changes, rebase to the latest `master` here and go ahead with the fix which is based already on that new `ConsumerProperties`.

And again: thank you for your contribution! Just a discussion driving us to the right direction is already a good contribution!

Walliee:
Created PR #283 for using ConsumerProperties. I'll rebase this PR once that gets merged.

Issue #282, opened 2019-08-19
KMS: Support async commits, groupId and clientId overrides


garyrussell:
See https://github.com/spring-projects/spring-kafka/issues/1203

The commit type (and callback) should be in the common consumer properties.

PR #283, opened 2019-08-21
Use ConsumerProperties in KafkaMessageSource

Walliee:


src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageSource.java
```diff
@@ -90,8 +92,6 @@
  */
 public class KafkaMessageSource<K, V> extends AbstractMessageSource<Object> implements Pausable {
 
-	private static final long DEFAULT_POLL_TIMEOUT = 50L;
```
Walliee: N.B. The default poll timeout value from `ConsumerProperties` is now 5000 ms. This is significantly higher than what it was previously. Consequently, the assign timeout value is now going to be 100 s.

Do we need to preserve the existing values?

Walliee:
There are still a few more properties that need to be populated using setters on `KafkaMessageSource` e.g. `messageConverter`, `payloadType`, `rawMessageHeader`. 

Does it make sense to create a configuration sub-class of `ConsumerProperties`?

Walliee:
This PR doesn't honor commit callback, async commits, commit action logging configuration from `ConsumerProperties`. These can be partially or fully handled in #282

Changes related to `topicPattern` and `topicPartitionsToAssign` will be covered as part of #281

src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageSource.java
garyrussell: Good point; I didn't think about that - I think we need a boolean on `ConsumerProperties` to indicate whether it was explicitly set, or defaulted. In the latter case we can override it to 50ms.

garyrussell:
>These can be partially or fully handled 

Agreed.

>Does it make sense to create a configuration sub-class of ConsumerProperties?

I think they can remain on the message source (we have a similar situation on the listener containers where a few properties are set there directly).

Let's see what @artembilan says.

src/main/java/org/springframework/integration/kafka/dsl/Kafka.java
```diff
@@ -74,72 +76,73 @@
...
-	 * @param topics the topic(s).
+	 * @param consumerProperties the consumerProperties.
 	 * @param <K> the Kafka message key type.
 	 * @param <V> the Kafka message value type.
 	 * @return the spec.
 	 * @since 3.0.1
 	 */
 	public static <K, V> KafkaInboundChannelAdapterSpec<K, V> inboundChannelAdapter(
-			ConsumerFactory<K, V> consumerFactory, String... topics) {
+			ConsumerFactory<K, V> consumerFactory, ConsumerProperties consumerProperties) {
```
artembilan: This is a breaking change. We could allow to do such a radical modification only in the next major version.
Right now we only have a choice to keep old ctos as `@Deprecated` and we may remove them in the next minor version then.

I know it is hard work to do in this code base, but we don't have a choice to provide a backward compatible code with a smooth migration option.

Thanks for udertstanding.

src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageSource.java
artembilan: And what is the problem to be consistent with listener container and rely on whatever is default in the `ConsumerProperties`?

src/main/java/org/springframework/integration/kafka/dsl/Kafka.java
Walliee: Okay, I am guessing we'd also want to maintain backwards compatibility with XSD syntax then?

artembilan: That’s correct assumption. With mentioning in the old attribute descriptions that they are deprecated and will be removed in the next version. Thank you again!

Walliee: Okay, I'll have an update soon. Is it okay to remove the required flag for `group-id` though?
Then in `KafkaInboundChannelAdapterParser` we can check if `consumer-factory` is set then use that otherwise use raw properties. 

https://github.com/spring-projects/spring-integration-kafka/blob/1903849cea7bf035a373309868d0fc8a2252f100/src/main/resources/org/springframework/integration/kafka/config/spring-integration-kafka-3.2.xsd#L179-L185

artembilan: Yes. It doesn’t look like breaking change to me

Walliee: Restored all pre-3.2 constructors and setters. Added depreciation warning to them.

src/main/java/org/springframework/integration/kafka/dsl/Kafka.java
```diff
@@ -80,66 +82,125 @@
 	 * @return the spec.
 	 * @since 3.0.1
 	 */
+	@Deprecated
```
artembilan: Please, also add a `@deprecated` tag into JavaDocs of the `@Deprecated` API

src/main/java/org/springframework/integration/kafka/dsl/Kafka.java
```diff
@@ -80,66 +82,125 @@
...
 	public static <K, V> KafkaInboundChannelAdapterSpec<K, V> inboundChannelAdapter(
 			ConsumerFactory<K, V> consumerFactory,
-			KafkaAckCallbackFactory<K, V> ackCallbackFactory, String... topics) {
+			KafkaAckCallbackFactory<K, V> ackCallbackFactory,
+			boolean allowMultiFetch,
+			String... topics) {
 
-		return inboundChannelAdapter(consumerFactory, ackCallbackFactory, false, topics);
+		return new KafkaInboundChannelAdapterSpec<>(consumerFactory, new ConsumerProperties(topics), ackCallbackFactory, allowMultiFetch);
```
artembilan: Please, consider do not exceed `120` line length.

src/main/java/org/springframework/integration/kafka/dsl/KafkaInboundChannelAdapterSpec.java
```diff
@@ -31,33 +32,51 @@
...
+	}
+
+	KafkaInboundChannelAdapterSpec(ConsumerFactory<K, V> consumerFactory,
+			ConsumerProperties consumerProperties,
+			KafkaAckCallbackFactory<K, V> ackCallbackFactory, boolean allowMultiFetch) {
+
+		this.target = new KafkaMessageSource<>(consumerFactory, consumerProperties, ackCallbackFactory, allowMultiFetch);
+	}
+
+	@Deprecated
```
artembilan: Please, also add JavaDocs into all the `public` API with the `@deprecated` tag and explanation in favor of what it is deprecated.

Thanks

src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageSource.java
```diff
@@ -153,10 +153,40 @@
...
+	@Deprecated
 	public KafkaMessageSource(ConsumerFactory<K, V> consumerFactory, String... topics) {
-		this(consumerFactory, new KafkaAckCallbackFactory<>(), false, topics);
+		this(consumerFactory, new ConsumerProperties(topics), new KafkaAckCallbackFactory<>(), false);
+	}
+
+	/**
+	 * Construct an instance with the supplied parameters. Fetching multiple
+	 * records per poll will be disabled.
+	 *
```
artembilan: No blank line in the method JavaDocs, please: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageSource.java
```diff
@@ -206,20 +240,36 @@ public KafkaMessageSource(ConsumerFactory<K, V> consumerFactory,
...
 	public KafkaMessageSource(ConsumerFactory<K, V> consumerFactory,
-			KafkaAckCallbackFactory<K, V> ackCallbackFactory, boolean allowMultiFetch, String... topics) {
+			ConsumerProperties consumerProperties,
+			KafkaAckCallbackFactory<K, V> ackCallbackFactory,
+			boolean allowMultiFetch) {
 
 		Assert.notNull(consumerFactory, "'consumerFactory' must not be null");
 		Assert.notNull(ackCallbackFactory, "'ackCallbackFactory' must not be null");
-		Assert.isTrue(topics != null && topics.length > 0, "At least one topic is required");
+		Assert.isTrue(consumerProperties.getTopics() != null && consumerProperties.getTopics().length > 0, "At least one topic is required");
```
artembilan: I hope you are going to address this in your next PR for `topicPattern` support...

src/test/kotlin/org/springframework/integration/kafka/dsl/KafkaDslKotlinTests.kt
```diff
@@ -57,10 +57,7 @@ import org.springframework.kafka.core.DefaultKafkaConsumerFactory
 import org.springframework.kafka.core.DefaultKafkaProducerFactory
 import org.springframework.kafka.core.KafkaTemplate
 import org.springframework.kafka.core.ProducerFactory
-import org.springframework.kafka.listener.ContainerProperties
-import org.springframework.kafka.listener.GenericMessageListenerContainer
-import org.springframework.kafka.listener.KafkaMessageListenerContainer
-import org.springframework.kafka.listener.MessageListenerContainer
+import org.springframework.kafka.listener.*
```
artembilan: No asterisk imports at all, please.

Walliee: Ah, my Kotlin imports were not set up.

src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageSource.java
Walliee: Yeah, I'll handle that as part of #281

src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageSource.java
Walliee: Okay, but I think the lint didn't warn about it.

artembilan: Well, we don't talk about some tool. It is just a matter of Spring best practice for code style.

Walliee:
@artembilan I incorporated your feedback. Let me know if it looks good now.

garyrussell:
LGTM, but we need a note in the what's new about the new default poll timeout - this can be a problem if they have several adapters and are using the default task scheduler which only has 10 threads since the thread will be tied up for 5 seconds if there are no new records.

artembilan:
> in the what's new about the new default poll timeout 


Right, but that's already a Spring Kafka project mattern. Doesn't look like we document anything here..

PR #287, opened 2019-09-30
Allow configuration of header-mapper in outbound-channel-adapter XML

tomvandenberge:
The XML configuration of an outbound-channel-adapter and outbound-gateway currently don't allow configuring a custom KafkaHeaderMapper. The KafkaProducerMessageHandler has a property for it, and it is already possible to set it using the DSL. It seems to be an omission in the XML configuration.

This pull request adds support for this.

### DIFF
--- a/src/test/java/org/springframework/integration/kafka/config/xml/KafkaOutboundGatewayParserTests.java
+++ b/src/test/java/org/springframework/integration/kafka/config/xml/KafkaOutboundGatewayParserTests.java
@@ -66,6 +66,8 @@ public class KafkaOutboundGatewayParserTests {
 				.isSameAs(this.context.getBean("failures"));
 		assertThat(TestUtils.getPropertyValue(this.messageHandler, "sendSuccessChannel"))
 				.isSameAs(this.context.getBean("successes"));
+		assertThat(TestUtils.getPropertyValue(this.messageHandler, "headerMapper"))
+		.isSameAs(this.context.getBean("customHeaderMapper"));
 	}
 
 	public static class EMS extends DefaultErrorMessageStrategy {
--- a/src/test/java/org/springframework/integration/kafka/config/xml/KafkaOutboundGatewayParserTests.java
+++ b/src/test/java/org/springframework/integration/kafka/config/xml/KafkaOutboundGatewayParserTests.java
@@ -31,6 +31,7 @@ import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
 
 /**
  * @author Gary Russell
+ * @author Tom van den Berge
  * @since 3.2
  *
  */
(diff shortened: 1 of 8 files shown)


src/test/java/org/springframework/integration/kafka/config/xml/KafkaOutboundGatewayParserTests.java
```diff
@@ -66,6 +66,8 @@ public void testProps() {
 				.isSameAs(this.context.getBean("failures"));
 		assertThat(TestUtils.getPropertyValue(this.messageHandler, "sendSuccessChannel"))
 				.isSameAs(this.context.getBean("successes"));
+		assertThat(TestUtils.getPropertyValue(this.messageHandler, "headerMapper"))
+		.isSameAs(this.context.getBean("customHeaderMapper"));
```
artembilan: Looks like wrong indention.

src/test/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandlerTests.java
```diff
@@ -329,6 +329,26 @@ public void testOutboundWithAsyncResults() {
 		producerFactory.destroy();
 	}
 
+	@Test
+	public void testOutboundWithCustomHeaderMapper() throws Exception {
+		DefaultKafkaProducerFactory<Integer, String> producerFactory = new DefaultKafkaProducerFactory<>(
```
artembilan: You should consider to `destroy()` this factory in the end of test.

src/test/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandlerTests.java
```diff
@@ -195,8 +196,6 @@ public void testOutbound() throws Exception {
 		assertThat(record).has(key(2));
 		assertThat(record).has(partition(1));
 		assertThat(record.value()).isNull();
-
-		producerFactory.destroy();
```
artembilan: Why have you removed this one? `Cut-and-paste` artefact? 😉

tomvandenberge:
Sure, no problem. Thanks for reviewing!

src/test/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandlerTests.java
tomvandenberge: Darn, this is stupid!

src/test/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandlerTests.java
```diff
@@ -195,7 +196,7 @@ public void testOutbound() throws Exception {
 		assertThat(record).has(key(2));
 		assertThat(record).has(partition(1));
 		assertThat(record.value()).isNull();
-
+
```
artembilan: I think these are wrong symbols for indentations.
Consider to use only *tabs*

tomvandenberge: This is becoming embarrassing ;-)
I hope everything is OK now.

tomvandenberge:
Great, I'm happy to be able to contribute a bit.

Will you merge this change to older versions (e.g. 2.3.x), too, or will it only be in the current one?

artembilan:
> this change to older versions 

No, we are not going to do that.

The `3.2` version of this project is going to be released today.

As a workaround for the version you are stuck, you always can use a `<service-activator>` with a reference to a `KafkaProducerMessageHandler` bean.

Thank you for contribution; looking forward for more!

PR #292, opened 2019-10-30
Apply SI-Kotlin-DSL in tests

artembilan:
* Upgrade dependencies


artembilan:
So, here is the fix for Kotlin tests based on the latest SI-Kotlin-DSL.

We may consider to merge this when that SI-Kotlin-DSL is released.

Thanks

artembilan:
I think after upgrading to the latest releases it is OK to merge.

Thanks

PR #310, opened 2021-11-17
Improve GRADLE build Performance

hongdongni:
[File system watching](https://blog.gradle.org/introducing-file-system-watching). Since Gradle 6.5, File system watching was introduced which can help to avoid unnecessary I/O. This feature is the default since 7.0. For an older version, we can enable this feature by setting `org.gradle.vfs.watch=true`.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,3 @@
 version=3.3.2.BUILD-SNAPSHOT
 org.gradle.caching=true
+org.gradle.vfs.watch = true


artembilan:
This project is out i ok f support because it version is in the end of life already for a long time .

we probably will consider to archive this repository …

garyrussell:
This is still used with SK 2.5 (and whatever cloud version uses that); they (Boot 2.3) goes EOL 2/15/22 so we should probably hold off until then.

